PR #542, opened 2025-06-27
[Certora] Exchange rate

gd-colin:
[Slack thread](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1758128802441769).

This PR shows that price of share variations are bound.

### DIFF
--- a/certora/applyMunging.patch
+++ b/certora/applyMunging.patch
@@ -17,7 +17,7 @@ diff -ruN VaultV2.sol VaultV2.sol
      function accrueInterest() public {
          (uint256 newTotalAssets, uint256 performanceFeeShares, uint256 managementFeeShares) = accrueInterestView();
 +
-+        lossRealization = lossRealization || newTotalAssets < _totalAssets;
++        lossRealization = newTotalAssets < _totalAssets;
 +
          emit EventsLib.AccrueInterest(_totalAssets, newTotalAssets, performanceFeeShares, managementFeeShares);
          _totalAssets = newTotalAssets.toUint128();


certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+
+import "Invariants.spec";
+
+// Duration of 10 years in seconds with 3 leap days.
+definition tenYears() returns uint256 = 60 * 60 * 24 * (3 + 365 * 10);
+
+// Check that the value of shares is rounded at most one share down upon deposit.
```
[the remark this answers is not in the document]
QGarchery: The same applies to other rules

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,150 @@
...
+
+    uint256 assetsBefore = currentContract._totalAssets;
+    uint256 supplyBefore = currentContract.totalSupply;
+
+    redeem(e, shares, receiver, onBehalf);
+
+    uint256 assetsAfter = currentContract._totalAssets;
+    uint256 supplyAfter = currentContract.totalSupply;
+
+    assert (assetsAfter + 1) * (supplyBefore + V) <= (assetsBefore + 1) * (supplyAfter + V) + (supplyBefore + V - 1);
```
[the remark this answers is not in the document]
QGarchery: I think we can find something similar for `mint`

certora/confs/ExchangeRate.conf
```diff
@@ -0,0 +1,34 @@
...
+    "-depth 12",
+    "-mediumTimeout 300",
+    "-timeout 300",
+    "-dontStopAtFirstSplitTimeout true",
+    "-enableStorageSplitting true",
+    "-smt_nonLinearArithmetic true",
+    "-smt_groundQuantifiers false",
+    "-smt_signedMulAxioms true",
+    "-smt_initialSplitDepth 3",
+    "-smt_usePz3 true"
```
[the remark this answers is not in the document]
colin-morpho: Right!? No, I checked out the source code to look for more prover flags.

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,150 @@
...
+    uint256 assetsBefore = currentContract._totalAssets;
+    uint256 supplyBefore = currentContract.totalSupply;
+
+    uint256 loss;
+    (loss, _) = realizeLoss(e, adapter, data);
+
+    uint256 assetsAfter = currentContract._totalAssets;
+    uint256 supplyAfter = currentContract.totalSupply;
+
+    assert loss > 0 => (assetsAfter + 1) * (supplyBefore + V) <= (assetsBefore + 1) * (supplyAfter + V);
```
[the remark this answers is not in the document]
colin-morpho: It does, fixed in https://github.com/morpho-org/vault-v2/pull/542/commits/2ca8df6a705a5d5a138870d445141a846d5dbe93

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,143 @@
...
+
+import "Invariants.spec";
+
+// Duration of 10 years in seconds.
+definition tenYears() returns uint256 = 60 * 60 * 24 * 365 * 10;
+
+// Check that deposit can raise the price-per-share by no more than burning one additional share would.
+rule sharePriceBoundDeposit(env e, uint256 assets, address onBehalf){
+    require (e.block.timestamp == currentContract.lastUpdate, "assume no interest is accrued");
+    require (currentContract.totalSupply > 0, "assume that the vault is seeded");
```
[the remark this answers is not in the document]
colin-morpho: Fixed in https://github.com/morpho-org/vault-v2/pull/542/commits/aafc82841048b8329b21fbd8e1af0c9be8d6782b.

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,143 @@
...
+
+    requireInvariant balanceOfZero();
+    requireInvariant totalSupplyIsSumOfBalances();
```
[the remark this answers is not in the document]
colin-morpho: Fixed in https://github.com/morpho-org/vault-v2/pull/542/commits/aafc82841048b8329b21fbd8e1af0c9be8d6782b.

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,143 @@
...
+
+    uint256 assetsBefore = currentContract._totalAssets;
+    uint256 supplyBefore = currentContract.totalSupply;
+
+    deposit(e, assets, onBehalf);
+
+    uint256 assetsAfter = currentContract._totalAssets;
+    uint256 supplyAfter = currentContract.totalSupply;
+
+    assert (assetsAfter + 1) * (supplyBefore + V - 1) <= (assetsBefore + 1) * (supplyAfter + V);
```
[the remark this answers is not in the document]
colin-morpho: Fixed by https://github.com/morpho-org/vault-v2/pull/542/commits/aafc82841048b8329b21fbd8e1af0c9be8d6782b and https://github.com/morpho-org/vault-v2/pull/542/commits/48dda9e8d4d3b24fc46c5e3175bbd8bde465daf4

.github/workflows/certora.yml
QGarchery, line 1: Marking it as draft because it doesn't go through yet

certora/confs/ExchangeRate.conf
```diff
@@ -0,0 +1,20 @@
...
+    "certora/helpers/Utils.sol",
+    "munged/VaultV2.sol",
+    "test/mocks/ERC20Mock.sol"
+  ],
+  "parametric_contracts": [
+    "ERC20Mock",
+    "VaultV2"
+  ],
+  "prover_args": [
+    "-s [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:def{randomSeed=8},z3:def{randomSeed=9},z3:def{randomSeed=10}]"
```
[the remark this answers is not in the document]
jhoenicke: This option points to solver instability.  Sometimes the solver z3 can solve the problem very fast and other times it runs for hours, never making any progress.  The trick is to run ten copies of the solver with different randomness and if one of them succeeds we can stop.

It would be good to check if this is necessary here, although I think with the ratio math used here the solvers are often unstable.

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,92 @@
...
+// Check that if deposit adds one more share to the user than it does, then the share price would decrease following a deposit.
+rule sharePriceBoundDeposit(env e, uint256 assets, address onBehalf){
+    require (e.block.timestamp == currentContract.lastUpdate, "assume no interest is accrued");
+
+    mathint assetsBefore = assets();
+    mathint sharesBefore = shares();
+
+    deposit(e, assets, onBehalf);
+
+    assert assets() * sharesBefore <= assetsBefore * (shares() + 1);
```
jhoenicke, line 19: I think this should also hold with `<` instead of `<=` and it would match the rule comment (and similarly for the other "Bound" rules).

Although, the rule comment itself is kind of confusing.  It says, if the protocol would give the user one share more, it would lose money (share price decreases). The real property the user is interested in is that the user gets at most one share less than the fair amount.  Both properties are equivalent to each other, though.  And since we need to avoid rational numbers, you can only express them indirectly as an integer disequality after multiplying with all the denominators.

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+
+import "Invariants.spec";
+
+definition shares() returns mathint = currentContract.totalSupply + currentContract.virtualShares;
+
+definition assets() returns mathint = currentContract._totalAssets + 1;
+
+// Check that if deposit adds one more share to the user than it does, then the share price would decrease following a deposit.
```
MathisGD, line 10: i find it more meaningful
```suggestion
// Show that the share price increase is bounded by a/(s(s+1)).
```

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,92 @@
...
+    requireInvariant virtualSharesBounds();
+
+    mathint assetsBefore = assets();
+    mathint sharesBefore = shares();
+
+    f(e, a);
+
+    require (!currentContract.lossRealization, "assume no loss realization");
+
+    assert assetsBefore * shares() <= assets() * sharesBefore;
```
MathisGD, line 91: here we would be missing a bug where deposit would make the share price decrease when there has been a loss realization in the accrueInterest. Couldn't we bound the decrease when there is a loss realization (we could record the value instead of a bool).

MathisGD: alternatively, we could accrue interest first and then call the functions (probably simpler)

certora/specs/ExchangeRate.spec
```diff
@@ -0,0 +1,92 @@
...
+    mathint sharesBefore = shares();
+
+    redeem(e, shares, receiver, onBehalf);
+
+    assert (assets() - 1) * sharesBefore < assetsBefore * shares();
+}
+
+// Check that loss realization decreases the share price.
+rule lossRealizationMonotonic(env e, address adapter, bytes data){
+    require (e.block.timestamp == currentContract.lastUpdate, "assume no interest is accrued");
```
MathisGD, line 60: why?

certora/specs/ExchangeRate.spec
QGarchery: I don't understand what this is saying, what are `a` and `s`, and why do we look at `a/(s(s+1))`.

To me the property we want to show here is that the protocol does the best it can for the user. So for a given amount of assets passed as inputs, the shares are increased as much as possible such that the share price still increases. It shows that the rounding is perfect in that sense (can't do better). Maybe there is a better way to phrase it ?

certora/specs/ExchangeRate.spec
QGarchery: Good point. About your ideas:
1. use an integer for the loss realization, I'll try that but maybe that will be too heavy for the prover
2. calling accrueInterest first is a mess because of the firstTotalAssets mecanism. Notably calling it a second time is not easy to see that it's a no-op, in case newTotalAssets was 0 the first time notably

certora/specs/ExchangeRate.spec
MathisGD: a=assets, s=shares. I quickly made the maths and I think that this is equivalent to what we are proving, but the property is clearer: the share price increase is not more than something small

certora/specs/ExchangeRate.spec
QGarchery: Removed now :white_check_mark:

certora/specs/ExchangeRate.spec
QGarchery: In the end 2 was not so bad, because we can use `firstTotalAssets != 0` to make so accruing interest the second time is more clearly a no-op

MathisGD (approved):
LGTM

PR #617, opened 2025-07-22
[Certora] Verif allocations

gd-colin:
This PR ensures allocations are updated as expected.

### DIFF
--- a/certora/specs/AllocationMorphoVaultV1Adapter.spec
+++ b/certora/specs/AllocationMorphoVaultV1Adapter.spec
@@ -16,9 +16,9 @@ methods {
     function MorphoVaultV1Adapter.allocation() external returns uint256 envfree;
 
     function _.allocate(bytes data, uint256 assets, bytes4 bs, address a) external with (env e)
-        => morphoVaultV1AdapterWrapperSummary(e, true, data, assets, bs, a) expect (bytes32[], uint256) ;
+        => morphoVaultV1AdapterWrapperSummary(e, true, data, assets, bs, a) expect (bytes32[], uint256);
     function _.deallocate(bytes data, uint256 assets, bytes4 bs, address a) external with (env e)
-        => morphoVaultV1AdapterWrapperSummary(e, false, data, assets, bs, a) expect (bytes32[], uint256) ;
+        => morphoVaultV1AdapterWrapperSummary(e, false, data, assets, bs, a) expect (bytes32[], uint256);
     function _.realizeLoss(bytes, bytes4, address) external => DISPATCHER(true);
 
     function _.borrowRate(MorphoHarness.MarketParams, MorphoHarness.Market) external => constantBorrowRate expect uint256;


certora/applyMetaMorphoMunging.patch
```diff
@@ -0,0 +1,143 @@
...
+@@ -585,7 +602,11 @@
+     /// @inheritdoc IERC4626
+     function totalAssets() public view override returns (uint256 assets) {
+         for (uint256 i; i < withdrawQueue.length; ++i) {
+-            assets += MORPHO.expectedSupplyAssets(_marketParams(withdrawQueue[i]), address(this));
++            uint256 supplyShares =
++                IMorphoHarness(MORPHO).position(_marketParams(withdrawQueue[i]).id(), address(this)).supplyShares;
++            (uint256 totalSupplyAssets, uint256 totalSupplyShares,,) =
++                MorphoBalancesLib.expectedMarketBalances(MORPHO, _marketParams(withdrawQueue[i]));
++            assets += supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares);
```
QGarchery, line 99: This essentially removes the reason why metamorpho was munged in the first place, why not just using Metamorpho directly (without munging it) ?

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,159 @@
...
+    // Trick to require that all the following addresses are different.
+    require (MorphoMarketV1 == 0x10, "ack");
+    require (MorphoMarketV1Adapter == 0x11, "ack");
+    require (currentContract == 0x12, "ack");
+
+    MorphoHarness.MarketParams marketParams;
+    MorphoHarness.Id marketId;
+    (marketParams, marketId) = Utils.morphoMarketV1MarketParams(data);
+
+    require (MorphoMarketV1Adapter.allocation(marketParams) <= MorphoMarketV1.totalSupplyAssets(marketId), "assume the adapter's allocation is less than or equal to the total assets");
```
QGarchery, line 148: It seems like we can remove this require, see [this run](https://prover.certora.com/output/66603/949c134839314d5db7e23a8759b67ce1?anonymousKey=4aa12d8e5eb11f5d28320e1cb0fb976369b611bf)

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,159 @@
...
+    require (MorphoMarketV1Adapter.shares(marketId) <= MorphoMarketV1.totalSupplyShares(marketId), "assume the adapter's shares are less than or equal to the total shares");
```
QGarchery, line 149: Same, see [this run](https://prover.certora.com/output/66603/0e1877abb7a948669cdae6bfcc93569c?anonymousKey=715d01bc3569880b277de17229203a3561bdf690)

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,159 @@
...
+
+    // Ensure the VaultV2 and Morpho contracts are properly linked to the adapter in the conf file.
+    assert MorphoMarketV1Adapter.parentVault == currentContract;
+    assert MorphoMarketV1Adapter.morpho == MorphoMarketV1;
+
+    deallocate(e, MorphoMarketV1Adapter, data, assets);
```
QGarchery, line 155: Actually can't we also remove this (de)allocate call ?

colin-morpho: I'll report on here ASAP.

colin-morpho: > I think we should actually verify that allocation is greater than expected assets after a (de)allocation. And then we can use that to prove that realizeLoss doesn't revert

In the general case this is not true, I think. What if there was a donation before de.allocate then we have `allocation` < `expected` don't we> and if we de.allocate 0 assets and interests are null, then `allocation` after < `expected` after. Or, we need to assume that there was no donation before, `allocation` before $\leq$ `expected` before, in which case we are showing that de.allocation can't "do" donations.

certora/specs/AllocationMorphoMarketV1Adapter.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/617/commits/d482fee52b8eb3f0866266b656d57aeed720e557

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,159 @@
...
+    // Ensure the VaultV2 and Morpho contracts are properly linked to the adapter in the conf file.
+    assert MorphoMarketV1Adapter.parentVault == currentContract;
+    assert MorphoMarketV1Adapter.morpho == MorphoMarketV1;
+
+    bytes32[] adapterIds = MorphoMarketV1Adapter.ids(marketParams);
+
+    deallocate(e, MorphoMarketV1Adapter, data, assets);
+
+    assert forall uint i. i < adapterIds.length => ghostAllocationAfter[adapterIds[i]] == ghostAllocationBefore[adapterIds[i]] + ghostInterest - assets;
+    assert forall bytes32 id. (forall uint i . i >= adapterIds.length || id != adapterIds[i]) => ghostAllocationAfter[id] == ghostAllocationBefore[id];
```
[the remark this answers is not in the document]
colin-morpho: You're right! changed in https://github.com/morpho-org/vault-v2/pull/617/commits/d482fee52b8eb3f0866266b656d57aeed720e557.

PR #618, opened 2025-07-22
onchain vic

MathisGD:


src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -75,6 +76,13 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
 
         uint256 interest = expectedSupplyAssets(marketParams, shares[marketId]).zeroFloorSub(allocation(marketParams));
 
+        if (shares[marketId] == 0) {
+            allMarketParams.push(marketParams);
+        } else {
+            allMarketParams[allMarketParams.length - 1] = marketParams;
+            allMarketParams.pop();
```
QGarchery, line 83: This removes the last element of the list, probably not what is intended

src/vic/OnchainVic.sol
QGarchery, line 1: I think we can remove the SingleMorphoVaultV1Vic. Using the onchain vic for this use case is slightly less optimized, but worth it I think

src/VaultV2.sol
```diff
@@ -358,7 +363,22 @@ contract VaultV2 is IVaultV2 {
 
     function setIsAdapter(address account, bool newIsAdapter) external {
```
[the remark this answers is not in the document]
MathisGD: optimizing the cost of getting the list would be interesting, but indeed we don't care about the cost of adding/removing an adapter

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -136,4 +144,19 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
         return supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares);
     }
+
+    function totalAssetsNoLoss() external view returns (uint256) {
+        uint256 totalAssets = 0;
+        for (uint256 i = 0; i < allMarketParams.length; i++) {
+            totalAssets += max(
+                expectedSupplyAssets(allMarketParams[i], shares[allMarketParams[i].id()]),
+                allocation(allMarketParams[i])
+            );
```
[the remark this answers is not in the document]
MathisGD: >with 0 allocation and 0 loss

so you mean it has x shares worth 0 (now)?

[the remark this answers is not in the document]
MathisGD: or the adapter could "forget" its shares if they are worth zero

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -94,6 +97,16 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
 
+        if (shares[marketId] == 0 && assets > 0) {
+            for (uint256 i = 0; i < allMarketParams.length; i++) {
+                if (Id.unwrap(allMarketParams[i].id()) == Id.unwrap(marketId)) {
+                    allMarketParams[i] = allMarketParams[allMarketParams.length - 1];
+                    allMarketParams.pop();
+                    break;
+                }
+            }
+        }
```
QGarchery, line 108: This should happen after shares are accounted with `shares[marketId] -= redeemedShares`, otherwise this will never remove a market (you can't deallocate an amount > 0 to a market that has no shares)

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -94,6 +97,16 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
         require(msg.sender == parentVault, NotAuthorized());
         require(marketParams.loanToken == asset, LoanAssetMismatch());
 
+        if (shares[marketId] == 0 && assets > 0) {
+            for (uint256 i = 0; i < allMarketParams.length; i++) {
```
QGarchery, line 108: The allocator could pass the index `i` in the data. This is also compatible with the liquidity adapter, as it could be in the liquidityData as well

QGarchery: There is the annoying part that if a market is removed, you may need to update the liquidityData

MathisGD: yes, exactly for this reason I thought that it was better this way

MathisGD: note that the added gas is not much, because most of the time you don't create/empty a market

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -74,6 +75,8 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
         require(msg.sender == parentVault, NotAuthorized());
         require(marketParams.loanToken == asset, LoanAssetMismatch());
 
+        if (shares[marketId] == 0 && assets > 0) allMarketParams.push(marketParams);
```
adhusson, line 78: If you allocate 1 asset that gets rounded to zero, then to remove the market from the `allMarketParams` array you will need to allocate more before you can withdraw. Otherwise when you try to deallocate 1 asset (to remove the market from the array) the withdraw will revert.

PR #619, opened 2025-07-23
Remove vault address lib

QGarchery:
Motivation:
- align with other factories which don't have AddressLib
- the use case "I wrote down the parameters, and I want the corresponding deployed contract" is still possible. Before and after you need a RPC call anyway: before it is to make sure that the contract is actually deployed, and after it is done by calling `vaultV2Factory.vaultV2()`
- simplifies the test, especially since we use dynamic linking and it was still causing issues sometimes

Merged into this PR:
- #623 
- #624


MathisGD:
putting as draft as it's a change to src

src/VaultV2Factory.sol
```diff
@@ -7,14 +7,16 @@ import {IVaultV2Factory} from "./interfaces/IVaultV2Factory.sol";
...
     /// @dev Returns the address of the deployed VaultV2.
     function createVaultV2(address owner, address asset, bytes32 salt) external returns (address) {
-        address vaultV2 = address(new VaultV2{salt: salt}(owner, asset));
+        address newVaultV2 = address(new VaultV2{salt: salt}(owner, asset));
 
-        isVaultV2[vaultV2] = true;
-        emit CreateVaultV2(owner, asset, vaultV2);
+        isVaultV2[newVaultV2] = true;
+        vaultV2[owner][asset][salt] = newVaultV2;
+        emit CreateVaultV2(owner, asset, newVaultV2);
```
MathisGD, line 18: why don't we log the salt already?

MathisGD: #623

src/VaultV2Factory.sol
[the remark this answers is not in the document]
QGarchery: Yes, but at the same time I don't think it really prevents a use case right ? You needed the salt before to compute the address

MathisGD (approved):

adhusson (approved):

PR #620, opened 2025-07-23
[Certora] Gate spec

colin-morpho:


certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    init_state axiom forall address user. ghostStatusCanReceiveAssets[user] == false;
+}
+
+persistent ghost mapping(address => bool) ghostBalanceChangeAllowed {
+    init_state axiom forall address user. ghostBalanceChangeAllowed[user] == true;
+}
+
+hook Sstore ERC20.balanceOf[KEY address user] uint256 newBalance (uint256 oldBalance) {
+    if (!canReceiveAssets(user)) {
+        ghostBalanceChangeAllowed[user] = oldBalance >= newBalance && ghostBalanceChangeAllowed[user];
```
QGarchery, line 57:
```suggestion
        assert oldBalance >= newBalance;
```
I find this clearer, because it removes one ghost variable, and also you have to understand how the hook works anyway (you can't just look at the spec)

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+
+    function _.supply(MorphoMarketV1Adapter.MarketParams, uint256, uint256, address, bytes) external => NONDET;
+    function _.withdraw(MorphoMarketV1Adapter.MarketParams, uint256, uint256, address, address) external => NONDET;
+    function _.deposit(uint256, address) external => NONDET;
+    function _.withdraw(uint256, address, address) external => NONDET;
+    function _.interestPerSecond(uint256, uint256) external => NONDET;
+}
+
+persistent ghost mapping(address => bool) ghostStatusCanSendShares {
+    init_state axiom forall address user. ghostStatusCanSendShares[user] == false;
```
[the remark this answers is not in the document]
colin-morpho: Axioms removed in https://github.com/morpho-org/vault-v2/pull/620/commits/5a803e5b57ea0ece7bb635c9d4b9c700f986394e.

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+
+    function _.allocate(bytes data, uint256 assets, bytes4, address) external => DISPATCHER(true);
+    function _.deallocate(bytes data, uint256 assets, bytes4, address) external => DISPATCHER(true);
+    function _.realizeLoss(bytes, bytes4, address) external => DISPATCHER(true);
+
+    function _.supply(MorphoMarketV1Adapter.MarketParams, uint256, uint256, address, bytes) external => NONDET;
+    function _.withdraw(MorphoMarketV1Adapter.MarketParams, uint256, uint256, address, address) external => NONDET;
+    function _.deposit(uint256, address) external => NONDET;
+    function _.withdraw(uint256, address, address) external => NONDET;
+    function _.interestPerSecond(uint256, uint256) external => NONDET;
```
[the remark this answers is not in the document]
colin-morpho: Outdated summary removed in https://github.com/morpho-org/vault-v2/pull/620/commits/f8bc061b2b1535efaa2966c1e13b9b254bbed8eb.

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    require(asset() == ERC20, "ack");
+
+    // Trick to require that all the following addresses are different.
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "require that the vault and the adapters are allowed to receive assets");
+    require (currentContract.liquidityAdapter == 0x0 || currentContract.liquidityAdapter == MorphoMarketV1Adapter || currentContract.liquidityAdapter == MorphoVaultV1Adapter);
```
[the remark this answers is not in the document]
colin-morpho: Doc improved in https://github.com/morpho-org/vault-v2/pull/620/commits/5a803e5b57ea0ece7bb635c9d4b9c700f986394e.

certora/confs/Gates.conf
```diff
@@ -0,0 +1,26 @@
...
+  "files": [
+    "lib/metamorpho/certora/dispatch/ERC20Standard.sol",
+    "src/adapters/MorphoMarketV1Adapter.sol",
+    "src/adapters/MorphoVaultV1Adapter.sol",
+    "src/VaultV2.sol"
+  ],
+  "link": [
+    "MorphoMarketV1Adapter:parentVault=VaultV2",
+    "MorphoVaultV1Adapter:parentVault=VaultV2",
+    "VaultV2:asset=ERC20Standard"
```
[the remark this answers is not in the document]
QGarchery: We have other implementations of ERC20 that do not have the `unchecked` issue that OZ has on `totalSupply`. So we can't dispatch but maybe we could have multiple conf files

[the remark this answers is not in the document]
QGarchery: Only 2 other implementations at the moment: ERC20USDT and ERC20NoRevert

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+}
+
+persistent ghost mapping(address => bool) ghostBalanceChangeAllowed {
+    init_state axiom forall address user. ghostBalanceChangeAllowed[user] == true;
+}
+
+hook Sstore ERC20.balanceOf[KEY address user] uint256 newBalance (uint256 oldBalance) {
+    if (!canReceiveAssets(user)) {
+        ghostBalanceChangeAllowed[user] = oldBalance >= newBalance && ghostBalanceChangeAllowed[user];
+    } else if (!canSendAssets(user)) {
```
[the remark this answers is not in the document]
QGarchery: Right, I was confused on the summary. But I still think this change is worth it, because there might be cases were we test both, since we don't mention anything about the "receive assets" part in the "send assets" spec for example. So notably the "send assets" spec do enforce that the user can't send assets, but it doesn't say if the user (or any other account) is able to receive assets

certora/specs/Gates.spec
```diff
@@ -0,0 +1,134 @@
...
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "require that the vault and the adapters are allowed to send assets");
+    require (currentContract.liquidityAdapter == 0x0 || currentContract.liquidityAdapter == MorphoMarketV1Adapter || currentContract.liquidityAdapter == MorphoVaultV1Adapter, "require that the liquidity adapter is unset or a known implementation");
+
+    require (!canSendAssets(user), "assume that the user under scrutiny can't send assets");
+    require (ghostBalanceChangeAllowed[user] == true, "setup the ghost state");
```
QGarchery, line 129:
```suggestion
    require (ghostBalanceChangeAllowed[user], "setup the ghost state");
```
(style) there are other occurrences

certora/specs/Gates.spec
```diff
@@ -0,0 +1,134 @@
...
+    require(currentContract.sendAssetsGate != 0, "require gating to be enabled for sending assets");
+
+    // Trick to require that all the following addresses are different.
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "require that the vault and the adapters are allowed to send assets");
+    require (currentContract.liquidityAdapter == 0x0 || currentContract.liquidityAdapter == MorphoMarketV1Adapter || currentContract.liquidityAdapter == MorphoVaultV1Adapter, "require that the liquidity adapter is unset or a known implementation");
```
QGarchery, line 126: It's a bit weird to state that actually, because the liquidity adapter can change at will by allocators. Could we require that the adapters are only one of the 2 adapter type that we developed instead ?

certora/specs/Gates.spec
```diff
@@ -0,0 +1,134 @@
...
+}{
+    require(currentContract.sendAssetsGate != 0, "require gating to be enabled for sending assets");
+
+    // Trick to require that all the following addresses are different.
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "require that the vault and the adapters are allowed to send assets");
```
QGarchery, line 125:
```suggestion
    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "do not check if the vault or the adapters themselves are properly gated to not send assets");
```
otherwise it's just stating what the require does. By the way this could be added as a description of this spec, something like "Check that no account (except the vault and the adapters) can send the underlying asset if the sendAssetsGate returns false for this account"

certora/specs/Gates.spec
```diff
@@ -0,0 +1,134 @@
...
+    f(e, args);
+
+    assert ghostBalanceChangeAllowed[user];
+}
+
+rule cantSendAssets(env e, method f, calldataarg args, address user)  filtered {
+    f -> f.selector != sig:MorphoMarketV1Adapter.skim(address).selector &&
+         f.selector != sig:MorphoVaultV1Adapter.skim(address).selector
+}{
+    require(currentContract.sendAssetsGate != 0, "require gating to be enabled for sending assets");
```
QGarchery, line 117:
```suggestion
    require(currentContract.sendAssetsGate != 0, "setup gating");
    require (!canSendAssets(user), "setup gating");
```
and remove the corresponding line below. This explains why we do those requires: we place ourselves in a scenario where a user is gated, and check that the corresponding change cannot be made

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    if (!canReceiveAssets(user)) {
+        ghostBalanceChangeAllowed[user] = oldBalance >= newBalance && ghostBalanceChangeAllowed[user];
+    } else if (!canSendAssets(user)) {
+        ghostBalanceChangeAllowed[user] = oldBalance <= newBalance && ghostBalanceChangeAllowed[user];
+    }
+}
+
+rule cantReceiveShares(env e, method f, calldataarg args, address user) filtered {
+    f -> f.selector != sig:MorphoMarketV1Adapter.skim(address).selector &&
+         f.selector != sig:MorphoVaultV1Adapter.skim(address).selector
```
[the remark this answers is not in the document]
QGarchery: Here this spec is on the shares, so it's different from the discussion linked. It really is a bug if you can transfer shares to someone who can't receive shares (even if it is done via the skim function)

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: Fixed in https://github.com/morpho-org/vault-v2/pull/620/commits/e6e65f964426024292842ca910c3fdecfef75ef7.

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: Fixed in https://github.com/morpho-org/vault-v2/pull/620/commits/e6e65f964426024292842ca910c3fdecfef75ef7.

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: Handled in https://github.com/morpho-org/vault-v2/pull/620/commits/e6e65f964426024292842ca910c3fdecfef75ef7.

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+
+using ERC20Standard as ERC20;
+using ERC20Helper as ERC20Helper;
+using MorphoMarketV1Adapter as MorphoMarketV1Adapter;
+using MorphoVaultV1Adapter as MorphoVaultV1Adapter;
+
+methods {
+    function multicall(bytes[]) external => NONDET DELETE;
```
QGarchery, line 10:
```suggestion
    function multicall(bytes[]) external => HAVOC_ALL DELETE;
```
since it's not a view function, and it should go through as it shouldn't be called

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+    function _.transferFrom(address, address, uint256) external => DISPATCHER(true);
+}
+
+function summaryMorphoMarketV1Supply(MorphoMarketV1Adapter.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns (uint256, uint256) {
+    assert shares == 0;
+    assert data.length == 0;
+    uint256 returnedShares;
+    require (ghostStatusCanReceiveAssets[MorphoMarketV1Adapter.morpho], "require that MorphoMarketV1 can receive assets");
+    ERC20Helper.safeTransferFrom(marketParams.loanToken, onBehalf, MorphoMarketV1Adapter.morpho, assets);
+    return (returnedShares, shares);
```
QGarchery, line 51:
```suggestion
    return (assets, returnedShares);
```

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+
+    uint256 sharesBefore = balanceOf(user);
+
+    f(e, args);
+
+    assert balanceOf(user) <= sharesBefore;
+}
+
+rule cantSendShares(env e, method f, calldataarg args, address user, uint256 shares) {
+    require(currentContract.sharesGate != 0, "require gating to be enabled for shares");
```
QGarchery, line 109:
```suggestion
    require(currentContract.sharesGate != 0, "setup gating");
    require (!canSendShares(user), "setup gating");
```
can also be done in the rule above

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+
+    function _.transfer(address, uint256) external => DISPATCHER(true);
+    function _.transferFrom(address, address, uint256) external => DISPATCHER(true);
+}
+
+function summaryMorphoMarketV1Supply(MorphoMarketV1Adapter.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns (uint256, uint256) {
+    assert shares == 0;
+    assert data.length == 0;
+    uint256 returnedShares;
+    require (ghostStatusCanReceiveAssets[MorphoMarketV1Adapter.morpho], "require that MorphoMarketV1 can receive assets");
```
QGarchery, line 49: The reason for this require is not very informative, this is an assumption we make right ?
It would be cleaner to put that in the list of accounts we don't check, that is adding it to the following line in each spec:
`require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract);`

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+    require (!canSendAssets(user), "setup gating");
+
+    // Trick to require that all the following addresses are different.
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "do not check if the vault or the adapters themselves are properly gated to not send assets");
+    require (forall address adapter . isAdapter(adapter) => (adapter == MorphoMarketV1Adapter || adapter == MorphoVaultV1Adapter), "require that the liquidity adapter is unset or a known implementation");
```
QGarchery, line 139: The reason is outdated, we could say something like "Only check the property for adapters that have been developed until now"

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+    require(currentContract.sendAssetsGate != 0, "setup gating");
+    require (!canSendAssets(user), "setup gating");
+
+    // Trick to require that all the following addresses are different.
+    require(MorphoMarketV1Adapter == 0x10, "ack");
+    require(MorphoVaultV1Adapter == 0x11, "ack");
+    require(currentContract == 0x12, "ack");
+    require(asset() == 0x13, "ack");
+
+    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "do not check if the vault or the adapters themselves are properly gated to not send assets");
```
QGarchery, line 138:
```suggestion
    require (user != MorphoMarketV1Adapter && user != MorphoVaultV1Adapter && user != currentContract, "do not check if the vault or the adapters themselves are properly gated to not send/receive assets");
```

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+    uint256 returnedShares;
+    require (ghostStatusCanSendAssets[MorphoMarketV1Adapter.morpho], "require that MorphoMarketV1 can send assets");
+    ERC20Helper.safeTransferFrom(marketParams.loanToken, MorphoMarketV1Adapter.morpho, onBehalf, assets);
+    return (assets, returnedShares);
+}
+
+function summaryMorphoVaultV1Deposit(uint256 assets, address receiver) returns uint256 {
+    uint256 shares;
+    require (ghostStatusCanReceiveAssets[MorphoMarketV1Adapter.morpho], "require that MorphoMarketV1 can receive assets");
+    ERC20Helper.safeTransferFrom(currentContract.asset, MorphoVaultV1Adapter, MorphoMarketV1Adapter.morpho, assets);
```
QGarchery, line 65: Wondering if we should not summarize it as 2 transfers: one from the adapter to the vault v1 and one from the vault v1 to market v1. This way we would catch the fact that vault v1 also sees its balance change

certora/specs/Gates.spec
```diff
@@ -0,0 +1,146 @@
...
+    require (!canSendShares(user), "require that the user under scrutiny can't send shares");
+
+    uint256 sharesBefore = balanceOf(user);
+
+    f(e, args);
+
+    assert balanceOf(user) >= sharesBefore;
+}
+
+rule cantSendAssetsAndCantReceiveAssets(env e, method f, calldataarg args, address user)  filtered {
```
QGarchery, line 125: Can we add a description of what each rule is checking ? I think it would be useful here because we make some assumptions in the verification

certora/specs/Gates.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/620/commits/59a8d61de8f995cd921530c5d47ab8140e15abcd.

certora/specs/Gates.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/620/commits/59a8d61de8f995cd921530c5d47ab8140e15abcd.

certora/specs/Gates.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/620/commits/59a8d61de8f995cd921530c5d47ab8140e15abcd.

certora/specs/Gates.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/620/commits/59a8d61de8f995cd921530c5d47ab8140e15abcd.

certora/specs/Gates.spec
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/620/commits/59a8d61de8f995cd921530c5d47ab8140e15abcd.

certora/specs/Gates.spec
colin-morpho: Reverted in https://github.com/morpho-org/vault-v2/pull/620/commits/c225e84583c8ff60e483dcb5bd32ec0220eeb3e8 as it makes the verification timeout, see [here](https://prover.certora.com/output/129828/69be18e9d8814fa4af6866c1f267f3ed?anonymousKey=6d4a40f3a15d3390ffce428729c8033be46f3c5f) and [here](https://prover.certora.com/output/129828/983433ff35d14dcbb63213a8563b8fa7?anonymousKey=cf048f7fa476e495fc9382794888ce8a55e079e1).

certora/specs/Gates.spec
```diff
@@ -0,0 +1,155 @@
...
+
+methods {
+    function asset() external returns address envfree;
+    function balanceOf(address) external returns uint256 envfree;
+    function canReceiveShares(address) external returns bool envfree;
+    function canSendShares(address) external returns bool envfree;
+    function canSendAssets(address) external returns bool envfree;
+    function canReceiveAssets(address) external returns bool envfree;
+    function isAdapter(address) external returns bool envfree;
+    function ERC20.totalSupply() external returns uint256 envfree;
```
colin-morpho, line 17:
```suggestion
```

colin-morpho:
Merged @QGarchery refactor with some observations in this https://github.com/morpho-org/vault-v2/pull/695#issuecomment-3178503644.

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    function _.deallocate(bytes data, uint256 assets, bytes4, address) external => DISPATCHER(true);
+    function _.realAssets() external => DISPATCHER(true);
+
+    function _.transfer(address, uint256) external => DISPATCHER(true);
+    function _.transferFrom(address, address, uint256) external => DISPATCHER(true);
+}
+
+function summaryMorphoMarketV1Supply(MorphoMarketV1Adapter.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) returns (uint256, uint256) {
+    assert shares == 0;
+    assert data.length == 0;
```
MathisGD, line 44: it's like nice to check but not needed right?

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    assert shares == 0;
+    uint256 returnedShares;
+    ERC20Helper.safeTransferFrom(marketParams.loanToken, MorphoMarketV1Adapter.morpho, onBehalf, assets);
+    return (assets, returnedShares);
+}
+
+function summaryMorphoVaultV1Deposit(uint256 assets, address receiver) returns uint256 {
+    uint256 shares;
+    ERC20Helper.safeTransferFrom(currentContract.asset, MorphoVaultV1Adapter, MorphoVaultV1Adapter.morphoVaultV1, assets);
+    ERC20Helper.safeTransferFrom(currentContract.asset, MorphoVaultV1Adapter.morphoVaultV1, MorphoMarketV1Adapter.morpho, assets);
```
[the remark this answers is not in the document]
colin-morpho: You mean why not one single transfer? See https://github.com/morpho-org/vault-v2/pull/620#discussion_r2263739540

certora/specs/Gates.spec
colin-morpho: Mmm, it sort of is. If we assume that there could be a callback, then it could be possible to change the asset's balance of a given user using another contract.

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    ERC20Helper.safeTransferFrom(currentContract.asset, MorphoMarketV1Adapter.morpho, MorphoVaultV1Adapter.morphoVaultV1, assets);
+    ERC20Helper.safeTransferFrom(currentContract.asset, MorphoVaultV1Adapter.morphoVaultV1, MorphoVaultV1Adapter, assets);
+    return shares;
+}
+
+persistent ghost mapping(address => bool) ghostCanSendShares;
+persistent ghost mapping(address => bool) ghostCanReceiveShares;
+persistent ghost mapping(address => bool) ghostCanSendAssets;
+persistent ghost mapping(address => bool) ghostCanReceiveAssets;
+persistent ghost mapping(address => bool) noInvalidBalanceChange;
```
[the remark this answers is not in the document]
colin-morpho: I kind of don't, the last rule would really weird to whether it's 

```suggestion
assert invalidBalanceChange[user];
```

or, negating the boolean value,

```suggestion
assert !invalidBalanceChange[user];
```

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+    require(currentContract == 0x12, "ack");
+
+    uint256 sharesBefore = balanceOf(user);
+
+    f(e, args);
+
+    assert balanceOf(user) >= sharesBefore;
+}
+
+// Check that transfers initiated from the vault, assuming the vault is not reentred, may only increase the balance of a given user when he can't send, and similarly the balance may only decrease when he can't receive.
```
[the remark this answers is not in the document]
colin-morpho: You can read `transfers initiated from the vault` as transfers following a vault operation, meaning that transfers deemed valid upon checking the gate while executing a vault operation, is that clearer?

certora/specs/Gates.spec
```diff
@@ -0,0 +1,149 @@
...
+persistent ghost mapping(address => bool) noInvalidBalanceChange;
+
+// A balance change is invalid if the balance increases when the user can't receive assets, or if it decreases when the user can't send assets.
+hook Sstore ERC20.balanceOf[KEY address user] uint256 newBalance (uint256 oldBalance) {
+    if (!canReceiveAssets(user) && newBalance > oldBalance) {
+        noInvalidBalanceChange[user] = false;
+    }
+    if (!canSendAssets(user) && newBalance < oldBalance) {
+        noInvalidBalanceChange[user] = false;
+    }
```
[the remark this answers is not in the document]
colin-morpho: You're right, that's why we inspect balances changes rather transfer calls!

certora/specs/Gates.spec
MathisGD: so let's remove that?

colin-morpho: To be more accurate, if we can't prove this assert then the summary is a bad approximation, it's like since we can assert that, it's fine to summarize like that.

certora/specs/Gates.spec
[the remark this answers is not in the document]
MathisGD: `f` could be a transfer on the erc20 no?

certora/specs/Gates.spec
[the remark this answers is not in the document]
MathisGD: but actually the vault has a harder spec (it prevents transfer 0 to a unauthorized address), so we could check transfers right?

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: You mean a user could call something like `IERC20(VaultV2.asset).transfer(X,Y)`? If so, this is not a vault operation, is it?

certora/specs/Gates.spec
MathisGD: I'm not sure to understand why?

certora/specs/Gates.spec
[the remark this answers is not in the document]
MathisGD: my initial question is: where do we prevent f to be a underlying token call

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: Mmm, yup we could I guess, but there many non practical points, I use a hook on `CALL` you can't inspect the arguments (not possible to decode `calldata`), using a summary is not the best because we have different implementation of ERC20 to pick from (namely for skimming) so we have to differentiate ERC20 transfer to `asset` and the skimmed token.

certora/specs/Gates.spec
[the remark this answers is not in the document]
colin-morpho: Because the configuration's parametric contract's do not apply to the underlying `asset`'s implementation contract, as it would make no sense.

certora/specs/Gates.spec
MathisGD: the prover would fine counter examples if there is a callback right?

colin-morpho: It should, but it's not as much as whether or not prover will find them, it's rather that we're making an approximation of that function and we're justifying why it's fine, we do it too in metamorpho see [here](https://github.com/morpho-org/metamorpho-v1.1/blob/ca1d9b2099391f6c333bf9a4b1d438804ab4fcf4/certora/specs/Tokens.spec#L32).

colin-morpho: We could just write comment for what is worth...

MathisGD (approved):

PR #621, opened 2025-07-23
[Certora] Import Chainsecurity specs

QGarchery:
[Slack thread](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1758135860344119).

This imports the verification done by Chainsecurity. Todo:
- [x] we need to make sure we can use it. Let's also mention that they did this work (in the README for example)
- [x] make sure that they go through
- [x] format it according to our conventions
- [x] decide which rule/invariant we want to keep

### DIFF
--- a/certora/specs/ChangesMorphoMarketV1Adapter.spec
+++ b/certora/specs/ChangesMorphoMarketV1Adapter.spec
@@ -42,8 +42,7 @@ rule changeForAllocateOrDeallocateIsBoundedByAllocation(env e, bytes data, uint2
     bytes32[] ids;
     int256 change;
 
-    bool isAllocate;
-    ids, change = allocateOrDeallocate(isAllocate, e, data, assets, selector, sender);
+    ids, change = allocateOrDeallocate(e, data, assets, selector, sender);
 
     assert allocation + change >= 0;
 }
--- a/certora/specs/ChangesMorphoVaultV1Adapter.spec
+++ b/certora/specs/ChangesMorphoVaultV1Adapter.spec
@@ -60,8 +60,7 @@ rule changeForAllocateOrDeallocateIsBoundedByAllocation(env e, bytes data, uint2
 
     bytes32[] ids;
     int256 change;
-    bool isAllocate;
-    ids, change = allocateOrDeallocate(isAllocate, e, data, assets, selector, sender);
+    ids, change = allocateOrDeallocate(e, data, assets, selector, sender);
 
     require vaultV1.balanceOf(currentContract) <= vaultV1.totalSupply(), "total supply is the sum of the balances";
 
--- a/certora/specs/IdsMorphoMarketV1Adapter.spec
+++ b/certora/specs/IdsMorphoMarketV1Adapter.spec
@@ -30,10 +30,9 @@ rule adapterAlwaysReturnsTheSameIDsForSameData(MorphoMarketV1Adapter.MarketParam
 // Show that the ids returned on allocate or deallocate match the reference id list.
 rule matchingIdsOnAllocateOrDeallocate(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
   MorphoMarketV1Adapter.MarketParams marketParams = Utils.decodeMarketParams(data);
-  bytes32[] ids;
 
-  bool isAllocate;
-  ids, _ = allocateOrDeallocate(isAllocate, e, data, assets, selector, sender);
+  bytes32[] ids;
+  ids, _ = allocateOrDeallocate(e, data, assets, selector, sender);
 
   bytes32[] idsMarket = ids(marketParams);
   assert idsMarket.length == 3;
--- a/certora/specs/IdsMorphoVaultV1Adapter.spec
+++ b/certora/specs/IdsMorphoVaultV1Adapter.spec
@@ -27,9 +27,7 @@ rule adapterAlwaysReturnsTheSameIDsForSameData() {
 // Show that the ids returned on allocate or deallocate match the reference id list.
 rule matchingIdsOnAllocateOrDeallocate(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
   bytes32[] ids;
-
-  bool isAllocate;
-  ids, _ = allocateOrDeallocate(isAllocate, e, data, assets, selector, sender);
+  ids, _ = allocateOrDeallocate(e, data, assets, selector, sender);
 
   bytes32[] idsAdapter = ids();
   assert idsAdapter.length == 1;
--- a/certora/specs/UtilityFunctions.spec
+++ b/certora/specs/UtilityFunctions.spec
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 // Copyright (c) 2025 Morpho Association
 
-function allocateOrDeallocate(bool isAllocate, env e, bytes data, uint256 assets, bytes4 selector, address sender) returns (bytes32[], int256) {
+function allocateOrDeallocate(env e, bytes data, uint256 assets, bytes4 selector, address sender) returns (bytes32[], int256) {
     bytes32[] ids;
     int256 change;
 
+    bool isAllocate;
     if (isAllocate) {
         ids, change = allocate(e, data, assets, selector, sender);
     } else {
--- a/test/ERC20Test.sol
+++ b/test/ERC20Test.sol
@@ -18,14 +18,11 @@ contract ERC20Test is BaseTest {
         uint256 deadline;
     }
 
-    function _signPermit(
-        uint256 privateKey,
-        address owner,
-        address to,
-        uint256 shares,
-        uint256 nonce,
-        uint256 deadline
-    ) internal view returns (uint8, bytes32, bytes32) {
+    function _signPermit(uint256 privateKey, address owner, address to, uint256 shares, uint256 nonce, uint256 deadline)
+        internal
+        view
+        returns (uint8, bytes32, bytes32)
+    {
         bytes32 hashStruct = keccak256(abi.encode(PERMIT_TYPEHASH, owner, to, shares, nonce, deadline));
         bytes32 digest = keccak256(abi.encodePacked("\x19\x01", vault.DOMAIN_SEPARATOR(), hashStruct));
         return vm.sign(privateKey, digest);


certora/specs/CS_MorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,262 @@
...
+    and thus calling f on the adapter only in adapterAlwaysReturnsTheSameIDsForSameData
+    is not sufficient to be sure ids won't change.
+  */
+  assert !queried_external;
+}
+
+/*
+  - ids() always return the same result for the same input data (in this case, input data is empty)
+*/
+rule adapterAlwaysReturnsTheSameIDsForSameData(method f) filtered {
```
colin-morpho, line 63:
```suggestion
rule adapterAlwaysReturnsTheSameIDsForSameData(env e, method f, calldataarg args) filtered {
```

certora/specs/CS_MorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,262 @@
...
+  f -> !f.isView
+} {
+  env e;
```
colin-morpho, line 66:
```suggestion
```

certora/specs/CS_MorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,262 @@
...
+  require(!queried_external);
+  require(adapter.morphoVaultV1 == metamorpho, "Very important otherwise the same tokens might be accounted for twice.");
+  require(metamorpho.MORPHO == morpho, "Fix morpho address.");
+  require(adapter.parentVault == vaultv2, "We know the parent vault.");
+  require(vaultv2.asset == metamorpho._asset, "We know the asset.");
+  require(vaultv2.sharesGate == 0x0000000000000000000000000000000000000000);
+
+  bytes32[] idsPre = adapter.ids();
+
+  calldataarg args;
```
colin-morpho, line 76:
```suggestion
```

colin-morpho (commented):
Style

certora/confs/CS_MorphoMarketV1Adapter.conf
```diff
@@ -0,0 +1,28 @@
...
+  "compiler_map": {
+    "Morpho": "solc-0.8.19",
+    "src/**/*.sol": "solc-0.8.28",
+    "test/**/*.sol": "solc-0.8.28",
+    "certora/**/*.sol": "solc-0.8.28"
+  },
+  "verify": "MorphoMarketV1Adapter:certora/specs/CS_MorphoMarketV1Adapter.spec",
+  "rule_sanity": "basic",
+  "server": "production",
+  "loop_iter": "2",
```
colin-morpho, line 23: This should be greater to 3, since the there's 3 ids in this adapter.

certora/specs/CS_MorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,256 @@
...
+  assert !queried_external;
+}
+
+/*
+  - ids() always return the same result for the same input data (market params)
+*/
+rule adapterAlwaysReturnsTheSameIDsForSameData(env e, method f, calldata args) filtered {
+  f -> !f.isView
+} {
+  require(vaultv2.sharesGate == 0x0000000000000000000000000000000000000000, "to avoid the canSendShares dispatch loop");
```
colin-morpho, line 69:
```suggestion
  require(vaultv2.sharesGate == 0, "to avoid the canSendShares dispatch loop");
```

certora/specs/AllocationVaultV2.spec
```diff
@@ -0,0 +1,76 @@
...
+    bytes32 id;
+    uint256 allocationPre = allocation(id);
+
+    accrueInterest(e);
+
+    uint256 allocationPost = allocation(id);
+
+    assert allocationPre == allocationPost;
+}
+
```
QGarchery, line 20:
```suggestion

```

certora/specs/ChangesMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,54 @@
...
+
+// Check that deallocate cannot return a change that would make the current allocation negative.
+rule changeForDeallocateIsBoundedByAllocation(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
+  Morpho.MarketParams marketParams = Utils.decodeMarketParams(data);
+  mathint allocation = allocation(marketParams);
+
+  bytes32[] ids; int256 change;
+  ids, change = deallocate(e, data, assets, selector, sender);
+
+  assert allocation + change >= 0;
```
colin-morpho, line 53: Ditto above

certora/specs/ChangesMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,54 @@
...
+
+// Check that allocate cannot return a change that would make the current allocation negative.
+rule changeForAllocateIsBoundedByAllocation(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
+  Morpho.MarketParams marketParams = Utils.decodeMarketParams(data);
+  mathint allocation = allocation(marketParams);
+
+  bytes32[] ids; int256 change;
+  ids, change = allocate(e, data, assets, selector, sender);
+
+  assert allocation + change >= 0;
```
colin-morpho, line 42: It would be nice to also show that casts are "safe".
```suggestion
  assert allocation + change >= 0;
  assert allocation + change == allocation(marketParams);
```

certora/specs/ChangesMorphoMarketV1Adapter.spec
colin-morpho: May be even merge the rule into single parametric one filtering only `allocate/deallocate`? I don't know if that is possible and I'm not really convinced it's the way to go.

certora/specs/ChangesMorphoVaultV1Adapter.spec
colin-morpho, line 1: Comments in market spec apply here as well.

certora/specs/ChangesMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,73 @@
...
+function mulDivSummary(uint256 x, uint256 y, uint256 denominator) returns uint256 {
+    mathint result;
+    if (denominator == 0) revert();
+    result = x * y / denominator;
+    if (result >= 2^256) revert();
+    return assert_uint256(result);
+}
+
+persistent ghost uint256 constantFeeShares;
+persistent ghost uint256 constantNewTotalAssets;
```
colin-morpho, line 31:
```suggestion
persistent ghost uint256 constantNewTotalAssets {
  // Market v1 stores assets on 128 bits, and there are at most 30 markets in vault v1
  axiom constantNewTotalAssets < 30 * 2^128;
}
```

certora/specs/ChangesMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,73 @@
...
+persistent ghost uint256 constantNewLostAssets;
+function constantAccrueFeeAndAssets() returns (uint256, uint256, uint256) {
+    require(constantNewTotalAssets < 30 * 2^128, "market v1 stores assets on 128 bits, and there are at most 30 markets in vault v1");
```
colin-morpho, line 34:
```suggestion
```

certora/specs/ChangesMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,73 @@
...
+    if (denominator == 0) revert();
+    result = x * y / denominator;
+    if (result >= 2^256) revert();
+    return assert_uint256(result);
+}
+
+persistent ghost uint256 constantFeeShares;
+persistent ghost uint256 constantNewTotalAssets;
+persistent ghost uint256 constantNewLostAssets;
+function constantAccrueFeeAndAssets() returns (uint256, uint256, uint256) {
```
colin-morpho, line 33: Could be changed to definition maybe? Optional

certora/specs/TotalAssetsChange.spec
colin-morpho, line 1: The rule total assets changed is quite hard to read, let's break it into many smaller lemmas. Then we can try to write a more general theorem by requiring the results from the lemmas.

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+    uint256 removedAssets;
+
+    if (f.selector == sig:deposit(uint,address).selector) {
+        require(removedAssets == 0, "this operation only adds assets");
+        deposit(e, addedAssets, e.msg.sender);
+    } else if (f.selector == sig:mint(uint,address).selector) {
+        uint256 shares;
+        require(removedAssets == 0, "this operation only adds assets");
+        require(addedAssets == previewMint(e, shares), "added assets should be the result of previewMint before the donation");
+        mint(e, shares, e.msg.sender);
```
colin-morpho, line 33:
```suggestion
        addedAssets = mint(e, shares, e.msg.sender);
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+    uint256 addedAssets;
+    uint256 removedAssets;
+
+    if (f.selector == sig:deposit(uint,address).selector) {
+        require(removedAssets == 0, "this operation only adds assets");
+        deposit(e, addedAssets, e.msg.sender);
+    } else if (f.selector == sig:mint(uint,address).selector) {
+        uint256 shares;
+        require(removedAssets == 0, "this operation only adds assets");
+        require(addedAssets == previewMint(e, shares), "added assets should be the result of previewMint before the donation");
```
colin-morpho, line 32:
```suggestion
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+        mint(e, shares, e.msg.sender);
+    } else if (f.selector == sig:withdraw(uint,address,address).selector) {
+        require(addedAssets == 0, "this operation only removes assets");
+        withdraw(e, removedAssets, e.msg.sender, e.msg.sender);
+    } else if (f.selector == sig:redeem(uint,address,address).selector) {
+        uint256 shares;
+        require(addedAssets == 0, "this operation only removes assets");
+        require(removedAssets == previewRedeem(e, shares), "redeemed assets should be the result of previewRedeem before the donation");
+        redeem(e, shares, e.msg.sender, e.msg.sender);
```
colin-morpho, line 41:
```suggestion
        removedAssets = redeem(e, shares, e.msg.sender, e.msg.sender);
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+        require(removedAssets == 0, "this operation only adds assets");
+        require(addedAssets == previewMint(e, shares), "added assets should be the result of previewMint before the donation");
+        mint(e, shares, e.msg.sender);
+    } else if (f.selector == sig:withdraw(uint,address,address).selector) {
+        require(addedAssets == 0, "this operation only removes assets");
+        withdraw(e, removedAssets, e.msg.sender, e.msg.sender);
+    } else if (f.selector == sig:redeem(uint,address,address).selector) {
+        uint256 shares;
+        require(addedAssets == 0, "this operation only removes assets");
+        require(removedAssets == previewRedeem(e, shares), "redeemed assets should be the result of previewRedeem before the donation");
```
colin-morpho, line 40:
```suggestion
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+        calldataarg args;
+        require(addedAssets == 0, "other operations don't add assets");
+        require(removedAssets == 0, "other operations don't remove assets");
+        f(e, args);
+    }
+
+    mathint totalAssetsPost = currentContract._totalAssets;
+
+    assert totalAssetsPre + addedAssets - removedAssets == totalAssetsPost;
```
colin-morpho, line 58:
```suggestion
    assert totalAssetsPre + addedAssets - removedAssets == currentContract._totalAssets;
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+        require(addedAssets == 0, "this operation only removes assets");
+        require(removedAssets == previewRedeem(e, shares), "redeemed assets should be the result of previewRedeem before the donation");
+        redeem(e, shares, e.msg.sender, e.msg.sender);
+    } else if (f.selector == sig:forceDeallocate(address,bytes,uint,address).selector) {
+        address adapter;
+        bytes data;
+        uint256 deallocationAmount;
+        require(addedAssets == 0, "this operation only removes assets");
+        require(removedAssets == mulDivUp(deallocationAmount, currentContract.forceDeallocatePenalty[adapter], 10^18), "compute the penalty quoted in assets");
+        forceDeallocate(e, adapter, data, deallocationAmount, e.msg.sender);
```
colin-morpho, line 48:
```suggestion
        removedAssets == forceDeallocate(e, adapter, data, deallocationAmount, e.msg.sender);
```

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,59 @@
...
+definition mulDivUp(uint256 x, uint256 y, uint256 z) returns mathint = (x * y + (z-1)) / z;
+
+// Check how total assets change without interest accrual.
+rule totalAssetsChange(env e, method f) filtered {
+    f -> !f.isView
+} {
+    mathint totalAssetsPre = currentContract._totalAssets;
+
+    uint256 addedAssets;
+    uint256 removedAssets;
```
colin-morpho, line 24:
```suggestion
    uint256 expectedAddedAssets;
    uint256 expectedrRemovedAssets;
```

While reading, I found the line `require(removedAssets == 0, "this operation only adds assets")` for the `deposit` case to be confusing, for instance I was wondering why isn't this being proved. But this is really the logical reflection of the computation of `_totalAssets`, so by the end it would be nice to be able to read something like expected vs. real.

certora/specs/AllocationVaultV2.spec
```diff
@@ -0,0 +1,64 @@
...
+    f(e, args);
+
+    uint256 allocationPost = allocation(id);
+
+    assert allocationPost != allocationPre => currentContract.liquidityAdapter != 0;
+}
+
+// Check that when no liquidity adapter is set, allocation can change via *allocate functions.
+rule functionsChangingAllocationWithoutLiquidityAdapter(env e, method f, calldataarg args) filtered {
```
MathisGD, line 48: not sure what this one is useful for? not the thing that we typically verify

certora/specs/ChangesMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,54 @@
...
+    function _.borrowRate(Morpho.MarketParams, Morpho.Market) external => constantBorrowRate expect uint256;
+    function _.borrowRateView(Morpho.MarketParams, Morpho.Market) external => constantBorrowRate expect uint256;
+
+    function Utils.decodeMarketParams(bytes) external returns (Morpho.MarketParams) envfree;
+}
+
+persistent ghost uint256 constantBorrowRate;
+
+// Check that calling allocate or deallocate with 0 amount yields the same change.
+rule sameChangeForAllocateAndDeallocateOnZeroAmount(env e, bytes data, bytes4 selector, address sender) {
```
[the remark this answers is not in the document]
QGarchery: Problem is that specs don't have the same assumptions, so if we create a third file then we end up duplicating the assumptions

certora/specs/IdsMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,45 @@
...
+using Utils as Utils;
+
+methods {
+    function ids() external returns (bytes32[]) envfree;
+
+    function Utils.havocAll() external envfree => HAVOC_ALL;
+}
+
+// Show that ids() is a constant function. It will be used as the reference id list in other rules.
+rule adapterAlwaysReturnsTheSameIDsForSameData() {
```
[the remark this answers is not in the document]
QGarchery: "the contract" is the vault adapter in this case.

Agree that the rule is not that useful, but I still like having it, as it gives a guideline for how this spec is supposed to work. Notably the corresponding version of this on the market adapter is more useful, and keeping it here makes so there is a framework for verifying this property if we add other adapters later

certora/specs/ChangesMorphoVaultV1Adapter.spec
QGarchery: I think all comments have been added now

certora/specs/TotalAssetsChange.spec
QGarchery: The idea is that we want to be able to predict how the state will change. Notice how for each case the function call is last

certora/helpers/Utils.sol
```diff
@@ -32,4 +33,16 @@ contract Utils {
...
+    function decodeMarketParams(bytes memory data) external pure returns (MarketParams memory) {
+        return abi.decode(data, (MarketParams));
+    }
+
+    function id(MarketParams memory marketParams) external pure returns (Id) {
+        return MarketParamsLib.id(marketParams);
+    }
+
+    function havocAll() external {
+        this.havocAll();
```
adhusson, line 46: is this to detect actual use of the function?

certora/specs/ChangesMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,54 @@
...
+  storage initialState = lastStorage;
+
+  bytes32[] idsAllocate; int256 changeAllocate;
```
adhusson, line 25:
```suggestion
  bytes32[] idsAllocate; 
  int256 changeAllocate;
```

certora/specs/ChangesMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,54 @@
...
+  idsAllocate, changeAllocate = allocate(e, data, 0, selector, sender) at initialState;
```
adhusson, line 26: at this point what is the difference between having `at initialState` and not having it?

adhusson (approved):

certora/helpers/Utils.sol
QGarchery: This is a helper function to summarize as `HAVOC_ALL`. The fact that it calls itself (indefinitely) is useful because in CVL direct calls to a function are not summarized. So, if in a rule you call this function, and you summarized it with `HAVOC_ALL`, it will:
- call the function directly (not summarize it)
- this function calls itself, which is considered and indirect call, and this time it will be summarized

certora/specs/ChangesMorphoMarketV1Adapter.spec
QGarchery: No difference, I removed it if it makes things clearer

certora/specs/UtilityFunctions.spec
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+
+function allocateOrDeallocate(bool isAllocate, env e, bytes data, uint256 assets, bytes4 selector, address sender) returns (bytes32[], int256) {
+    bytes32[] ids;
```
MathisGD: nitpick: could be like that
```suggestion
function allocateOrDeallocate(env e, bytes data, uint256 assets, bytes4 selector, address sender) returns (bytes32[], int256) {
    bool isAllocate;
    bytes32[] ids;
```

certora/specs/IdsMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,44 @@
...
+
+  assert idsPre.length == idsPost.length;
+  assert idsPre.length == 3;
+  assert idsPre[0] == idsPost[0];
+  assert idsPre[1] == idsPost[1];
+  assert idsPre[2] == idsPost[2];
+}
+
+// Show that the ids returned on allocate or deallocate match the reference id list.
+rule matchingIdsOnAllocateOrDeallocate(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
```
MathisGD:
```suggestion
rule allocateDeallocateIdsConsistency(env e, bytes data, uint256 assets, bytes4 selector, address sender) {
```

certora/specs/IdsMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,44 @@
...
+
+methods {
+    function ids(MorphoMarketV1Adapter.MarketParams) external returns (bytes32[]) envfree;
+
+    function Utils.havocAll() external envfree => HAVOC_ALL;
+    function Utils.decodeMarketParams(bytes) external returns(MorphoMarketV1Adapter.MarketParams) envfree;
+}
+
+// Show that ids() is a function that only depend on its input. It will be used as the reference id list in other rules.
+rule adapterAlwaysReturnsTheSameIDsForSameData(MorphoMarketV1Adapter.MarketParams marketParams) {
```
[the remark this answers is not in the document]
QGarchery: I don't like words like "consistent", "correct", "sane", "sound" when describing verification work as I feel like it hides what we are trying to verify. Each of those word can mean a number of different things, for example here it could also mean that they return data in the correct form (correct types), that they revert at appropriate times, or that they match an predefined order on the ids (sorted for example, or from most general to less general)

certora/specs/IdsMorphoMarketV1Adapter.spec
QGarchery: see https://github.com/morpho-org/vault-v2/pull/621#discussion_r2538848068

certora/specs/IdsMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,38 @@
...
```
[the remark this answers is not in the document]
QGarchery: see https://github.com/morpho-org/vault-v2/pull/621#discussion_r2538848068

certora/specs/TotalAssetsChange.spec
```diff
@@ -0,0 +1,86 @@
...
+methods {
+    function multicall(bytes[]) external => HAVOC_ALL DELETE;
+
+    function accrueInterestView() internal returns (uint256, uint256, uint256) => summaryAccrueInterestView();
+}
+
+// Assume that accrueInterest does nothing.
+function summaryAccrueInterestView() returns (uint256, uint256, uint256) {
+    return (currentContract._totalAssets, 0, 0);
+}
```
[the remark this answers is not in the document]
QGarchery: Not sure it would add much, because we can already assume that the starting state is the state with interest added to the computation. But it would be a bit annoying in the rules below because we should replace every `currentContract._totalAssets` occurences, to add interest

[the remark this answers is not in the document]
QGarchery: Yes not a huge deal I agree, just a some slight changes to the rules, but I don't feel like it adds much. This is because we already consider all the possible starting values of `_totalAssets`, so considering all possible values of `_totalAssets + interest` is similar. It would be a different story with shares minted to fee recipients, but this is precisely what we want to avoid with the "no interest accrual" assumption

PR #626, opened 2025-07-28
competition 3: clarify rate capping

adhusson:
addresses https://cantina.xyz/code/523e1540-f8c3-45ae-9c5d-b6d35d3a326c/findings/3

alternative to https://github.com/morpho-org/vault-v2/pull/650

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -30,6 +30,7 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The Vic might not distribute as much interest as planned if:
 /// - The Vic reverted on `setVic`.
 /// - The Vic returned an interest per second that is too high (it is capped at a max rate).
+/// - The interest is set to zero if the returned rate is more than the max rate.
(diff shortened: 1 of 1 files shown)


src/VaultV2.sol
```diff
@@ -30,6 +30,7 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The Vic might not distribute as much interest as planned if:
 /// - The Vic reverted on `setVic`.
 /// - The Vic returned an interest per second that is too high (it is capped at a max rate).
+/// - The interest is set to zero if the returned rate is more than the max rate.
```
MathisGD, line 33:
```suggestion
/// - The interest is set to zero if the returned interest per second corresponds to a rate greater than the max rate.
```

QGarchery, line 33:
```suggestion
/// - The Vic returned an interest per second that is too high (in which case the rate is set to the max rate).
```

QGarchery (approved):

MathisGD (changes requested):
I prefer #650

PR #628, opened 2025-07-28
cap realization incentive to half of assets

adhusson:


src/libraries/MathLib.sol
```diff
@@ -33,4 +33,11 @@ library MathLib {
         require(x <= type(uint128).max, ErrorsLib.CastOverflow());
         return uint128(x);
     }
+
+    /// @dev Returns the min of `x` and `y`.
+    function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        assembly {
+            z := xor(x, mul(xor(x, y), lt(y, x)))
+        }
+    }
```
QGarchery, line 42: We can add a test of this function in MathTest

QGarchery (approved):

MathisGD:
note: not needed if we go for #655

adhusson:
> note: not needed if we go for #655

why?

adhusson:
closing since #655 is go

PR #632, opened 2025-07-28
document interest per second rounding impact

adhusson:
* https://cantina.xyz/code/523e1540-f8c3-45ae-9c5d-b6d35d3a326c/findings/146

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ For now only a Vic of this type is provided, the [ManualVic](./src/vic/ManualVic
 - the Vic has an additional internal notion of max interest per second, to ensure that the role of allocator can be given more safely.
   The curator controls this internal notion of max interest per second, while the sentinels are only able to decrease it to reduce the risk of having a rate too high.
 
+The impact of rounding should be kept in mind and the interest per second should be adapter accordingly.
+The impact of rounding on the interest is higher for tokens with a high per-unit value, for smaller vaults, and for lower interest rates.
+For instance, 2% APR on 10e3 USDC is ~6.34 USDC / second. The round down to 6 reduces the APR by 5%.
+
 ### Bad debt realization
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,6 +93,10 @@ For now only a Vic of this type is provided, the [ManualVic](./src/vic/ManualVic
 - the Vic has an additional internal notion of max interest per second, to ensure that the role of allocator can be given more safely.
   The curator controls this internal notion of max interest per second, while the sentinels are only able to decrease it to reduce the risk of having a rate too high.
 
+The impact of rounding should be kept in mind and the interest per second should be adapter accordingly.
```
QGarchery, line 96:
```suggestion
The impact of rounding should be kept in mind and the interest per second should be adapted accordingly.
```

README.md
```diff
@@ -93,6 +93,10 @@ For now only a Vic of this type is provided, the [ManualVic](./src/vic/ManualVic
...
+The impact of rounding on the interest is higher for tokens with a high per-unit value, for smaller vaults, and for lower interest rates.
+For instance, 2% APR on 10e3 USDC is ~6.34 USDC / second. The round down to 6 reduces the APR by 5%.
```
QGarchery, line 98:
```suggestion
For instance, 2% APR on 10e3 USDC is ~6.34 USDC / second.
The round down to 6 reduces the APR by 5%.
```

README.md
QGarchery, line 1: I think one important aspect is that rounding is not only on interest (which is already documented [here](https://github.com/morpho-org/vault-v2/blob/da667f27d7b386bf159e3bc43de41e29af9c62a6/src/VaultV2.sol#L36-L37)), but also on fees, and it should be documented as well. So maybe only doing #627 is enough

PR #634, opened 2025-07-28
docs: liquidity market cap (competition 207)

MathisGD:
fixes https://cantina.xyz/code/523e1540-f8c3-45ae-9c5d-b6d35d3a326c/findings/207


src/VaultV2.sol
```diff
@@ -90,6 +90,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The liquidity adapter is useful on exit, so that exit liquidity is available in addition to the idle assets. But
 /// the same adapter/data is used for both entry and exit to have the property that in the general case looping
 /// supply-withdraw or withdraw-supply should not change the allocation.
+/// @dev If a cap (absolute or relative) associated with the ids returned by the liquidity adapter on the liquidity data
+/// is reached, deposit/mint will revert. So typically their caps should be high.
```
QGarchery, line 94: I'm worried that this doesn't make it clear enough that an issue like 207 is acknowledged: even with caps that are high, they could be lower than what an attacker is able to deposit to make a tx revert

adhusson (approved):

PR #648, opened 2025-07-29
Fix relative cap check (comp 53)

QGarchery:
See:
- https://morpholabs.slack.com/archives/C08A40KKN9W/p1753364945363219
- https://cantina.xyz/code/523e1540-f8c3-45ae-9c5d-b6d35d3a326c/findings/53

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -41,7 +41,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// _totalAssets that is passed.
 ///
 /// FIRST TOTAL ASSETS
-/// @dev The variable firstTotalAssets tracks the total assets after the first interest accrual of the transaction.
+/// @dev The variable underestimatedTotalAssets tracks the total assets after the first interest accrual of the
+/// transaction.
 /// @dev Used to implement a mechanism that prevents bypassing relative caps with flashloans.
 /// @dev This mechanism can generate false positives on relative cap breach when such a cap is nearly reached,
 /// for big deposits that go through the liquidity adapter.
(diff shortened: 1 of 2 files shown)


adhusson (approved):
Should add tests later

src/VaultV2.sol
```diff
@@ -56,7 +56,7 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// corresponding markets, and losses are deducted only when realized for these markets.
 /// @dev The caps are checked on allocate (where allocations can increase) for the ids returned by the adapter.
 /// @dev Relative caps are "soft" in the sense that they are only checked on allocate.
-/// @dev The relative cap is relative to totalAssets, or more precisely to firstTotalAssets.
+/// @dev The relative cap is relative to totalAssets, or more precisely to underestimatedTotalAssets.
```
MathisGD, line 59:
```suggestion
/// @dev The relative cap is relative to underestimatedTotalAssets.
```
people won't get it, I think

src/VaultV2.sol
```diff
@@ -40,8 +40,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The Vic must not call totalAssets() because it will try to accrue interest, but instead use the argument
 /// _totalAssets that is passed.
 ///
-/// FIRST TOTAL ASSETS
-/// @dev The variable firstTotalAssets tracks the total assets after the first interest accrual of the transaction.
+/// UNDERESTIMATED TOTAL ASSETS
+/// @dev The variable underestimatedTotalAssets is an underestimation of the current value of `_totalAssets`..
```
MathisGD, line 44: should we explain how it behaves?

QGarchery: I tried something, let me know what you think

src/VaultV2.sol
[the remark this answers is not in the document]
QGarchery: Agree that this PR is not very satisfying. But there is something to fix I think, because the following spec for the relative cap check in `allocate` is very natural: "anytime `allocate` is called, you can't allocate more than the relative cap, relative to the value of `_totalAssets` at that time". Else what is the spec of the relative cap check actually ?

PR #649, opened 2025-07-29
test: fix some tests

MathisGD:


test/AccrueInterestTest.sol
```diff
@@ -153,14 +190,13 @@ contract AccrueInterestTest is BaseTest {
         vm.stopPrank();
         vault.setPerformanceFee(performanceFee);
         vault.setManagementFee(managementFee);
-        vm.warp(vm.getBlockTimestamp() + elapsed);
-
-        // Rate too high.
         vm.prank(allocator);
         vic.setInterestPerSecondAndDeadline(interestPerSecond, type(uint64).max);
-        uint256 totalAssetsBefore = vault.totalAssets();
```
MathisGD, line 161: nothing was tested here because of this. I removed the use of totalAssets everywhere in the tests, and wrote a test for it specifically

test/AccrueInterestTest.sol
```diff
@@ -153,14 +190,13 @@ contract AccrueInterestTest is BaseTest {
         vm.stopPrank();
         vault.setPerformanceFee(performanceFee);
         vault.setManagementFee(managementFee);
-        vm.warp(vm.getBlockTimestamp() + elapsed);
```
MathisGD, line 156: warp really doesn't work, so I replaced it everywhere

test/AccrueInterestTest.sol
adhusson: I was looking at this right now, I think that nothing was tested also because in `vic.setInterestPerSecondAndDeadline`, `accrueInterest` is called (with the elapsed skip it should be good now).

adhusson (approved):

test/AccrueInterestTest.sol
```diff
@@ -141,9 +177,10 @@ contract AccrueInterestTest is BaseTest {
         performanceFee = bound(performanceFee, 0, MAX_PERFORMANCE_FEE);
         managementFee = bound(managementFee, 0, MAX_MANAGEMENT_FEE);
         deposit = bound(deposit, 0, MAX_TEST_ASSETS);
-        vm.assume(deposit.mulDivDown(MAX_RATE_PER_SECOND, WAD) <= type(uint96).max);
-        interestPerSecond = bound(interestPerSecond, deposit.mulDivDown(MAX_RATE_PER_SECOND, WAD), type(uint96).max);
-        elapsed = bound(elapsed, 0, 20 * 365 days);
+        elapsed = bound(elapsed, 1, 20 * 365 days);
+        vm.assume(interestPerSecond <= type(uint256).max / elapsed);
```
QGarchery, line 181:
```suggestion
        interestPerSecond = bound(interestPerSecond, 0, type(uint256).max / elapsed);
```

PR #655, opened 2025-07-29
solution 3: loop

MathisGD:
spec:
- totalAssets = Σ adapters[i].totalAssets(); we don't need a VIC anymore, this is handled directly by adapters.
- loss is realized automatically (and without incentive)
- totalAssets growth is capped at max rate, naturally handling donation and penalties. It's also possible to "fix the rate" with this.
- adapters return `int change` to update allocations (including losses)
- we don't need to prevent donations on markets in which the vault invests, because it doesn't increase the loss if someone makes a donation just before a realization. So we simplified the accounting in adapters.
- fixes the issue where the loss isn't precisely calculated because of interest distributed via the VIC but not accounted in allocations


src/VaultV2.sol
```diff
@@ -557,30 +550,28 @@ contract VaultV2 is IVaultV2 {
     /* EXCHANGE RATE FUNCTIONS */
 
     function accrueInterest() public {
-        if (lastUpdate != block.timestamp) {
```
MathisGD, line 560: would it be fine to do re-add that

test/RealizeLossTest.sol
```diff
@@ -27,166 +27,4 @@ contract RealizeLossTest is BaseTest {
...
-
-        // Realize the loss.
-        vm.prank(allocator);
-        vault.realizeLoss(address(adapter), hex"");
-        assertEq(
-            vault.allocation(expectedIds[0]), deposit - expectedLoss, "allocation should have decreased by the loss"
-        );
-    }
-
-    function testRealizeMoreThanTotalAssets(uint256 deposit, uint256 expectedLoss) public {
```
MathisGD, line 175: re-add

test/MorphoVaultV1AdapterTest.sol
```diff
@@ -184,119 +184,6 @@ contract MorphoVaultV1AdapterTest is Test {
...
-        deal(address(asset), address(adapter), deposit);
-        parentVault.allocateMocked(address(adapter), hex"", deposit);
-
-        // Realize loss.
-        (bytes32[] memory ids, uint256 loss) = parentVault.realizeLossMocked(address(adapter), hex"");
-        assertEq(ids, expectedIds, "ids");
-        assertEq(loss, 0, "loss");
-    }
-
-    function testLossRealization(uint256 deposit, uint256 _loss) public {
```
MathisGD, line 219: re-add

src/VaultV2.sol
MathisGD, line 1: should we remove the accrueInterest from allocate? would be great for gas

lib/morpho-blue
QGarchery, line 1: was morpho-blue outdated ?

src/VaultV2.sol
```diff
@@ -556,30 +560,30 @@ contract VaultV2 is IVaultV2 {
...
-        uint256 newTotalAssets = _totalAssets + interest;
+    function accrueInterestView() public view returns (uint256, uint256, uint256, uint256) {
+        uint256 realAssets = IERC20(asset).balanceOf(address(this));
+        for (uint256 i = 0; i < adapters.length; i++) {
+            realAssets += IAdapter(adapters[i]).totalAssets();
+        }
+        uint256 loss = _totalAssets.zeroFloorSub(realAssets);
+        uint256 maxInterest = (_totalAssets * (block.timestamp - lastUpdate)).mulDivDown(maxRate, WAD);
+        uint256 interest = MathLib.min(realAssets.zeroFloorSub(_totalAssets), maxInterest);
+        uint256 newTotalAssets = _totalAssets + interest - loss;
```
QGarchery, line 586:
```suggestion
        uint256 newTotalAssets;
        uint256 interest;
        if (realAssets >= _totalAssets) {
            uint256 maxInterest = (_totalAssets * (block.timestamp - lastUpdate)).mulDivDown(maxRate, WAD);
            interest = MathLib.min(realAssets - _totalAssets, maxInterest);
            newTotalAssets = _totalAssets + interest;
        } else {
            newTotalAssets = realAssets;
        }
```
not sure which is clearer

src/VaultV2.sol
```diff
@@ -556,30 +560,30 @@ contract VaultV2 is IVaultV2 {
...
-            _totalAssets = newTotalAssets.toUint192();
-            if (performanceFeeShares != 0) createShares(performanceFeeRecipient, performanceFeeShares);
-            if (managementFeeShares != 0) createShares(managementFeeRecipient, managementFeeShares);
-            lastUpdate = uint64(block.timestamp);
-        }
+        (uint256 newTotalAssets, uint256 loss, uint256 performanceFeeShares, uint256 managementFeeShares) =
+            accrueInterestView();
+        emit EventsLib.AccrueInterest(_totalAssets, newTotalAssets, performanceFeeShares, managementFeeShares);
+        _totalAssets = newTotalAssets.toUint128();
+        if (loss > 0) enterBlocked = true;
```
QGarchery, line 567: `loss == _totalAssets.zeroFlooSub(newTotalAssets)` (clearer with the refactor suggested in the comment above) so maybe we don't have to do the change where `accrueInterestView()` also returns the `loss`

src/VaultV2.sol
```diff
@@ -556,30 +560,30 @@ contract VaultV2 is IVaultV2 {
...
         if (firstTotalAssets == 0) firstTotalAssets = _totalAssets;
     }
 
     /// @dev Returns newTotalAssets, performanceFeeShares, managementFeeShares.
-    /// @dev Reverts if the call to the Vic reverts.
     /// @dev The management fee is not bound to the interest, so it can make the share price go down.
     /// @dev The performance and management fees are taken even if the vault incurs some losses.
     /// @dev Both fees are rounded down, so fee recipients could receive less than expected.
-    function accrueInterestView() public view returns (uint256, uint256, uint256) {
-        uint256 elapsed = block.timestamp - lastUpdate;
```
QGarchery, line 576: (optional) we can keep this variable I think, since it's clear what it means and it's used twice

lib/morpho-blue
MathisGD: I'm not sure what happened

src/VaultV2.sol
MathisGD: accrueInterestView should only return the enterBlocked

lib/morpho-blue
```diff
@@ -1 +1 @@
-Subproject commit d89ca53ff6cbbacf8717a8ce819ee58f49bcc592
+Subproject commit fcb190b0e0e1355defe56b19781ac573986c3f74
```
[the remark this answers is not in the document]
MathisGD: idk what happened tbh, I can roll back

src/libraries/ConstantsLib.sol
```diff
@@ -6,7 +6,7 @@ uint256 constant WAD = 1e18;
 bytes32 constant DOMAIN_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
 bytes32 constant PERMIT_TYPEHASH =
     keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
-uint256 constant MAX_RATE_PER_SECOND = 200e16 / uint256(365 days); // 200% APR
+uint256 constant MAX_MAX_RATE = 200e16 / uint256(365 days); // 200% APR
```
MathisGD, line 9: could be higher now

src/VaultV2.sol
```diff
@@ -563,38 +566,37 @@ contract VaultV2 is IVaultV2 {
...
         uint256 elapsed = block.timestamp - lastUpdate;
-        if (elapsed == 0) return (_totalAssets, 0, 0);
-
-        uint256 tentativeInterest = vic != address(0) ? IVic(vic).interest(_totalAssets, elapsed) : 0;
-        uint256 maxInterest = (_totalAssets * elapsed).mulDivDown(MAX_RATE_PER_SECOND, WAD);
-        uint256 interest = MathLib.min(tentativeInterest, maxInterest);
-        uint256 newTotalAssets = _totalAssets + interest;
+        uint256 realAssets = IERC20(asset).balanceOf(address(this));
+        for (uint256 i = 0; i < adapters.length; i++) {
+            realAssets += IAdapter(adapters[i]).totalAssets();
```
QGarchery, line 587: Should we rename it (`realAssets`, `ownedAssets` ?) to state that it's not ERC-4626 `totalAssets` ? The big difference is that the vault v2 might not be the only depositor of this ERC-4626 (that's why we use previewRedeem in Vault v1 adapter)

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -64,54 +65,54 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
-    {
+    /// @dev Returns the ids of the allocation and the change in allocation.
+    function allocate(bytes memory data, uint256 assets, bytes4, address) external returns (bytes32[] memory, int256) {
         MarketParams memory marketParams = abi.decode(data, (MarketParams));
         Id marketId = marketParams.id();
         require(msg.sender == parentVault, NotAuthorized());
         require(marketParams.loanToken == asset, LoanAssetMismatch());
 
-        uint256 interest = expectedSupplyAssets(marketParams, shares[marketId]).zeroFloorSub(allocation(marketParams));
+        if (shares[marketId] == 0 && assets > 0) allMarketParams.push(marketParams);
```
QGarchery, line 75: Here you could introduce duplicates if `allocate` is called multiple times and each time the assets round down to 0 shares

QGarchery:
See also #671

src/adapters/MorphoMarketV1Adapter.sol
MathisGD: I'm not sure that allocating allows you to deallocate all shares btw, you still get the rounding

src/VaultV2.sol
```diff
@@ -539,16 +542,16 @@ contract VaultV2 is IVaultV2 {
...
         require(isAdapter[adapter], ErrorsLib.NotAdapter());
 
-        (bytes32[] memory ids, uint256 interest) = IAdapter(adapter).deallocate(data, assets, msg.sig, msg.sender);
+        (bytes32[] memory ids, int256 change) = IAdapter(adapter).deallocate(data, assets, msg.sig, msg.sender);
 
         for (uint256 i; i < ids.length; i++) {
             Caps storage _caps = caps[ids[i]];
             require(_caps.allocation > 0, ErrorsLib.ZeroAllocation());
-            _caps.allocation = _caps.allocation + interest - assets;
+            _caps.allocation = uint256(int256(_caps.allocation) + change);
```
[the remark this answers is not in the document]
adhusson: either revert or zeroFloorAddInt, underflowing silently is not ok.

src/adapters/MorphoMarketV1Adapter.sol
MathisGD: fixed by #668

src/VaultV2.sol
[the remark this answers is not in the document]
QGarchery: Let's discuss in https://github.com/morpho-org/vault-v2/pull/673

PR #656, opened 2025-07-30
Docs interest instead of interest per second

QGarchery:


src/vic/ManualVic.sol
```diff
@@ -62,6 +62,7 @@ contract ManualVic is IManualVic {
     }
 
     /// @dev Returns the interest per second.
+    /// @dev One asset per second might already be above the max rate for assets with high value per unit.
```
MathisGD, line 65:
```suggestion
    /// @dev One asset of interest might already be above the max rate for assets with high value per unit.
```

src/VaultV2.sol
```diff
@@ -24,18 +24,17 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// initial deposit. See https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
 ///
 /// INTEREST / VIC
-/// @dev To accrue interest, the vault queries the Vault Interest Controller (Vic) which returns the interest per second
-/// that must be distributed over the period (since lastUpdate).
+/// @dev To accrue interest, the vault queries the Vault Interest Controller (Vic) which returns the interest that must
+/// be distributed over the period (since lastUpdate).
```
MathisGD, line 28:
```suggestion
/// be distributed.
```

MathisGD (approved):

PR #657, opened 2025-07-30
chore: separate fmt ci

MathisGD:
it was run for each decimal


adhusson (approved):
ok to add but not worth the added lines & added job I think, fmt is ~instant

QGarchery (changes requested):
Pretty sure it's more costly to after this PR, since adding a new job is a huge overhead (just downloading every submodule is taking way more time)

PR #658, opened 2025-07-30
solution 6: resync

adhusson:
done on top of onchain vic https://github.com/morpho-org/vault-v2/pull/618

- resync loops through all markets
- allocation is now real assets and used only for caps & market interaction protection
- add max rate per second to onchain vic

todo
- [x] Re-add realize loss tests
- [x] Add max rate per second tests
- [x] Check verif


src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -65,53 +67,59 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
-    }
+        uint256 currentSupply = expectedSupplyAssets(marketParams, shares[marketId]);
+        uint256 currentAllocation = allocation(marketParams);
 
-    function realizeLoss(bytes memory data, bytes4, address) external view returns (bytes32[] memory, uint256) {
-        MarketParams memory marketParams = abi.decode(data, (MarketParams));
-        require(marketParams.loanToken == asset, LoanAssetMismatch());
-
-        uint256 loss = allocation(marketParams) - expectedSupplyAssets(marketParams, shares[marketParams.id()]);
+        if (currentAllocation > 0 && currentSupply == 0) {
```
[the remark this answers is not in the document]
adhusson: If the adapter has 1 share it really has 2 assets. So it you withdraw 2 assets the `currentSupply` will be 0.

src/VaultV2.sol
```diff
@@ -512,11 +532,11 @@ contract VaultV2 is IVaultV2 {
         accrueInterest();
 
         SafeERC20Lib.safeTransfer(asset, adapter, assets);
-        (bytes32[] memory ids, uint256 interest) = IAdapter(adapter).allocate(data, assets, msg.sig, msg.sender);
+        (bytes32[] memory ids, int256 change) = IAdapter(adapter).allocate(data, assets, msg.sig, msg.sender);
 
         for (uint256 i; i < ids.length; i++) {
             Caps storage _caps = caps[ids[i]];
-            _caps.allocation = _caps.allocation + interest + assets;
+            _caps.allocation = _caps.allocation.zeroFloorAddInt(change);
```
MathisGD, line 539: why zero floor here?

src/VaultV2.sol
```diff
@@ -763,17 +783,17 @@ contract VaultV2 is IVaultV2 {
     /// losses do not cause issues.
     /// @dev The incentive will be null if the msg.sender isn't allowed to receive shares.
     /// @dev Returns incentiveShares, loss.
-    function realizeLoss(address adapter, bytes memory data) external returns (uint256, uint256) {
-        require(isAdapter[adapter], ErrorsLib.NotAdapter());
-
+    function resync() external returns (uint256, uint256) {
```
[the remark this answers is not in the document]
adhusson: renamed to `realizeLosses`

src/VaultV2.sol
MathisGD: yes but not more than allocation, right?

src/VaultV2.sol
```diff
@@ -762,18 +782,17 @@ contract VaultV2 is IVaultV2 {
     /// @dev For small losses, the incentive might be small, or even null because of rounding. This is ok as small
     /// losses do not cause issues.
     /// @dev The incentive will be null if the msg.sender isn't allowed to receive shares.
-    /// @dev Returns incentiveShares, loss.
-    function realizeLoss(address adapter, bytes memory data) external returns (uint256, uint256) {
-        require(isAdapter[adapter], ErrorsLib.NotAdapter());
-
+    function realizeLosses() external returns (uint256, uint256) {
```
MathisGD, line 785: Actually something like "clamp" would be better because it could well be that there was no loss (just too much interest has been distributed), unlike before

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -35,6 +35,8 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
     address public skimRecipient;
     /// @dev `shares` are the recorded shares created by allocate and burned by deallocate.
     mapping(Id => uint256) public shares;
+    Id[] public marketArray;
+    mapping(Id => uint256) public marketMapping;
```
MathisGD, line 39: this is to be more efficient upon removal right? why do we care about this? it shouldn't happen much no?

src/VaultV2.sol
adhusson: actually not with current adapters, my bad

src/VaultV2.sol
adhusson: but this will create a loss on the withdrawable amount for some depositors / on their accounting of how much their shares are worth

MathisGD: right, totalAsset loss but not necessarily realAsset loss

adhusson:
closing in favor of #655

PR #672, opened 2025-08-04
chore: update adapters interface

### LINKED ISSUE #667: adapter interfaces are outdated

MathisGD:
fixes #667


src/adapters/interfaces/IMorphoMarketV1Adapter.sol
QGarchery, line 1: I think we are still missing some functions. Here I see that `asset` and `adapterId` are public in the contract but are not in this interface

QGarchery (approved):

PR #673, opened 2025-08-04
Allocation update: floor at zero

QGarchery:
See https://github.com/morpho-org/vault-v2/pull/655#discussion_r2251541340


src/VaultV2.sol
QGarchery, line 1: Let's add tests later, see #674

adhusson:
Added https://github.com/morpho-org/vault-v2/pull/676 for completeness but prefer this one

PR #676, opened 2025-08-04
allocation update: revert on underflow

adhusson:
alternative to #673


src/VaultV2.sol
adhusson, line 1: Also adding tests later if needed

QGarchery (approved):

PR #680, opened 2025-08-05
Refactor hardcoded id

QGarchery:
To avoid being hardcoded twice, making it clear that it is the same value without having to check that it is the same characters in "this/marketParams" notably
Low prio


src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -113,18 +113,22 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
     function ids(MarketParams memory marketParams) public view returns (bytes32[] memory) {
         bytes32[] memory ids_ = new bytes32[](3);
         ids_[0] = adapterId;
         ids_[1] = keccak256(abi.encode("collateralToken", marketParams.collateralToken));
-        ids_[2] = keccak256(abi.encode("this/marketParams", address(this), marketParams));
+        ids_[2] = lastId(marketParams);
         return ids_;
     }
 
+    function lastId(MarketParams memory marketParams) internal view returns (bytes32) {
```
MathisGD, line 128: public?

PR #682, opened 2025-08-05
docs: reverting market

### LINKED ISSUE #664: reverting market in 3/6

MathisGD:
fixes #664


QGarchery (approved):
Also, if we have a market that reverts (constantly), there is the following mitigation:
- deallocate every other market of that adapter to idle
- remove the adapter

It's not perfect because you get no yield during that time, but at the same time it should happen rarely

PR #686, opened 2025-08-08
[Certora] Allocations

colin-morpho:
Replaces #617


certora/confs/AllocationMorphoMarketV1Adapter.conf
```diff
@@ -0,0 +1,39 @@
...
+    "MorphoHarness": "solc-0.8.19",
+    "MorphoMarketV1Adapter": "solc-0.8.28",
+    "Utils": "solc-0.8.28",
+    "VaultV2": "solc-0.8.28"
+  },
+  "optimistic_hashing": true,
+  "optimistic_loop": true,
+  "loop_iter": "4",
+  "allow_solidity_calls_in_quantifiers": true,
+  "exclude_rule": "allocationIsInt256",
```
colin-morpho, line 24: Informational, disabled as it would fail without deleting `multicall`.
Checked in `Invariants.conf`. Here only to make it possible to require in other rules.

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,159 @@
...
+
+    require (forall uint256 i. forall uint256 j. i < j && j < ids.length => ids[j] != ids[i], "assume that all returned ids are unique");
+    ghostChange = change;
+
+    return (ids, change);
+}
+
+definition max_int256() returns int256 = (2 ^ 255) - 1;
+
+strong invariant allocationIsInt256(bytes32 id)
```
colin-morpho, line 53: Informational: this definition is here only to make it possible to require in other rules. Importing `Invariants.spec` makes the process fail for a yet to be explained reason.

colin-morpho (commented):
Notes about this PR.

certora/specs/AllocationMorphoMarketV1Adapter.spec
colin-morpho: I suggest we track this in an issue and merge it as is until Certora comes up a fix/explanation.

certora/helpers/Utils.sol
```diff
@@ -36,4 +46,41 @@ contract Utils {
...
+
+    function expectedTotalSupplyShares(address morpho, MarketParams memory marketParams)
+        external
+        view
+        returns (uint256)
+    {
+        (, uint256 totalSupplyShares,,) = MorphoBalancesLib.expectedMarketBalances(IMorpho(morpho), marketParams);
+
+        return totalSupplyShares;
+    }
```
colin-morpho, line 74:
```suggestion
```

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,157 @@
...
+
+    function _.borrowRate(MorphoHarness.MarketParams, MorphoHarness.Market) external => constantBorrowRate expect uint256;
+    function _.borrowRateView(MorphoHarness.MarketParams, MorphoHarness.Market) external => constantBorrowRate expect uint256;
+
+    function _.allocate(bytes data, uint256 assets, bytes4 bs, address a) external with (env e)
+        => morphoMarketV1AdapterWrapperSummary(e, true, data, assets, bs, a) expect (bytes32[], int256);
+    function _.deallocate(bytes data, uint256 assets, bytes4 bs, address a) external with (env e)
+        => morphoMarketV1AdapterWrapperSummary(e, false, data, assets, bs, a) expect (bytes32[], int256);
+
+    function _.market(MorphoHarness.Id) external => DISPATCHER(true);
```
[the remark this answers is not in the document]
colin-morpho: It's called by `MorphoBalancesLib`, I think it's a similar case to the `SafeERC20Lib` case you mentioned in Telegram. The fact that the called address is "abstract" in the call site is not making the resolution possible.

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,157 @@
...
+    MorphoHarness.Id marketId;
+    (marketParams, marketId) = Utils.morphoMarketV1MarketParams(data);
+
+    allocate(e, MorphoMarketV1Adapter, data, assets);
+
+    uint256 allocation = MorphoMarketV1Adapter.allocation(marketParams);
+    uint256 supplyShares = MorphoMarketV1.position_(marketId, MorphoMarketV1Adapter).supplyShares;
+    uint256 expected = Utils.expectedSupplyAssets(e, MorphoMarketV1, marketParams, supplyShares);
+
+    assert allocation >= expected;
```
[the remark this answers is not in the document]
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/686/commits/f2af2d62006d6d18b8eb4d66b9f6d22f026225e0.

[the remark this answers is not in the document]
colin-morpho: Also this should hold for vault v1 adapter `allocate`, but not `deallocate` I think

certora/applyMetaMorphoMunging.patch
QGarchery, line 1: Can we make this patch minimal ? I see many changes that we don't need, like the `withdrawRank`. Do we need to change the interface as well to `IMorphoHarnes`, now that we don't call `expectedSupplyAssets` directly on Morpho

certora/specs/AllocationMorphoVaultV1Adapter.spec
```diff
@@ -0,0 +1,153 @@
...
+    bytes32 id;
+    uint256 allocationBefore = allocation(id);
+
+    requireInvariant allocationIsInt256(adapterIds[i]);
+    int256 idIAllocationBefore = assert_int256(allocation(adapterIds[i]));
+
+    allocate(e, MorphoVaultV1Adapter, data, assets);
+
+    assert allocation(adapterIds[i]) == idIAllocationBefore + ghostChange;
+    assert !(exists uint j . j < adapterIds.length && id == adapterIds[j]) => currentContract.caps[id].allocation == allocationBefore;
```
QGarchery, line 95:
```suggestion
    assert currentContract.caps[id].allocation != allocationBefore => exists uint j. j < adapterIds.length && id == adapterIds[j];
```
[optional] I find it a bit clearer, "only changes are in the id list"

QGarchery (commented):
Looks good 💯

certora/applyMetaMorphoMunging.patch
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/686/commits/1b4be5a82b5990702935dffd2ab41a0e99dd9590.

certora/specs/AllocationMorphoMarketV1Adapter.spec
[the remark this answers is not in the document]
colin-morpho: Done in https://github.com/morpho-org/vault-v2/pull/686/commits/5d1020f6538554041d8aa3994865ba6d13887c9d.

certora/applyMetaMorphoMunging.patch
colin-morpho: Fixed in https://github.com/morpho-org/vault-v2/pull/686/commits/0ec79aca28e4b6ebc6676ec4bf288394038c95f6, turns out that removing `IMophoHarness` breaks compilation due to reading `MORPHO.lastUpdate` line 273 of munged MetaMorpho.

QGarchery: There is no need for it in the original source code, so I don't think we need it either (considering that we didn't change this line). Isn't it fixed by adding back the `using MorphoBalancesLib for IMorpho;` line ?

colin-morpho: Right, but then it will not circumvent extSload which is the whole point of this munging, right?

QGarchery: Oh right good point, that makes sense. What about munging that line to change `MORPHO.lastUpdate(id)` into `MORPHO.market(id).lastUpdate` ? Avoiding `IMorphoHarness` seems worth it, as it is a big conceptual change

colin-morpho: Not convincing really, in fact there are many lines touched by this change namely 346, 622, 623, 712, 755, 787, 837, 839 you can see the changed in https://github.com/morpho-org/vault-v2/pull/686/commits/8b46ad4e233d26118509357b389c1093e4f0c436. I personally find it very invasive but I see your point on not using `IMorphoHarness` at all.

QGarchery (approved):

certora/applyMetaMorphoMunging.patch
```diff
@@ -0,0 +1,106 @@
...
+ import {MarketParamsLib} from "../lib/morpho-blue/src/libraries/MarketParamsLib.sol";
+ import {IERC20Metadata} from "../lib/openzeppelin-contracts/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+ import {MorphoBalancesLib} from "../lib/morpho-blue/src/libraries/periphery/MorphoBalancesLib.sol";
+@@ -45,9 +44,7 @@
+     using UtilsLib for uint256;
+     using SafeCast for uint256;
+     using SafeERC20 for IERC20;
+-    using MorphoLib for IMorpho;
+     using SharesMathLib for uint256;
+-    using MorphoBalancesLib for IMorpho;
```
QGarchery, line 18: We can probably reintroduce that one to minimize the diff. Note that we still would need to avoid to call `expectedSupplyAssets` as the goal of the munging here is to avoid calling `extSloads`

QGarchery:
To update, and also we could probably remove munging:
- on vault v2, see [this thread](https://morpholabs.slack.com/archives/C09EAFEJUMV/p1760719408891409)
- on metamorpho, see how it's done in #781

certora/specs/Invariants.spec
```diff
@@ -67,6 +70,9 @@ strong invariant decreaseTimelockTimelock()
 strong invariant totalSupplyIsSumOfBalances()
     totalSupply() == sumOfBalances;
 
+strong invariant allocationIsInt256(bytes32 id)
+    allocation(id) <= max_int256();
+
```
MathisGD:
```suggestion

strong invariant relativeCapBound(bytes32 id)
    relativeCap(id) <= 10^18;

strong invariant maxRateBound()
    maxRate() <= Utils.maxMaxRate();
```

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,129 @@
...
+}
+
+rule allocateChangesAdapterIds(env e, bytes data, uint256 assets) {
+    // Trick to require that all the following addresses are different.
+    require MorphoMarketV1 == 0x10, "ack";
+    require MorphoMarketV1Adapter == 0x11, "ack";
+    require currentContract == 0x12, "ack";
+
+    MorphoHarness.MarketParams marketParams = Utils.decodeMarketParams(data);
+    bytes32[] adapterIds = MorphoMarketV1Adapter.ids(marketParams);
```
MathisGD: adapterId is misleading
```suggestion
    bytes32[] resIds = MorphoMarketV1Adapter.ids(marketParams);
```

certora/specs/AllocationMorphoMarketV1Adapter.spec
```diff
@@ -0,0 +1,129 @@
...
+
+    uint i;
+    require i < adapterIds.length, "require i to be a valid index";
+    requireInvariant allocationIsInt256(adapterIds[i]);
+    int256 idIAllocationBefore = assert_int256(allocation(adapterIds[i]));
+
+    allocate(e, MorphoMarketV1Adapter, data, assets);
+
+    assert allocation(adapterIds[i]) == idIAllocationBefore + ghostChange;
+    assert !(exists uint j. j < adapterIds.length && id == adapterIds[j]) => currentContract.caps[id].allocation == allocationBefore;
```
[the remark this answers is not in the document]
QGarchery: Yes agree, see [here](https://github.com/morpho-org/vault-v2/pull/686#discussion_r2269799767)

PR #704, opened 2025-08-20
[Certora] relative caps are not violated

colin-morpho:
This PR:
  - check relative caps are not violated assuming no deallocation or that the relative cap is decreased;
  - check registered adapters are unique.

### DIFF
--- a/certora/specs/Invariants.spec
+++ b/certora/specs/Invariants.spec
@@ -71,3 +71,18 @@ strong invariant decreaseTimelockTimelock()
 
 strong invariant totalSupplyIsSumOfBalances()
     totalSupply() == sumOfBalances;
+
+strong invariant registeredAdaptersAreSet()
+    (forall uint256 i. i < currentContract.adapters.length => currentContract.isAdapter[currentContract.adapters[i]])
+{
+    preserved {
+        requireInvariant adaptersUnique();
+    }
+}
+strong invariant adaptersUnique()
+    forall uint256 i. forall uint256 j. (i < j && j < currentContract.adapters.length) => currentContract.adapters[j] != currentContract.adapters[i]
+{
+    preserved {
+        requireInvariant registeredAdaptersAreSet();
+    }
+}
--- a/certora/specs/RelativeCaps.spec
+++ b/certora/specs/RelativeCaps.spec
@@ -1,0 +1,39 @@
+// Spdx-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+
+using Utils as Utils;
+
+methods {
+    function multicall(bytes[]) external => NONDET DELETE;
+    function deallocateInternal(address, bytes memory, uint256) internal returns (bytes32[] memory) => deallocateInternalSummary();
+    function Utils.wad() external returns uint256 envfree;
+}
+
+persistent ghost bool deallocateInternalNotCalled {
+    init_state axiom deallocateInternalNotCalled;
+}
+
+function deallocateInternalSummary() returns (bytes32[]) {
+    deallocateInternalNotCalled = false;
+
+    bytes32[] ids;
+    uint i;
+    havoc currentContract.caps[ids[i]].allocation assuming i < ids.length ;
+
+    return ids;
+}
+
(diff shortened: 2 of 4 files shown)


lilCertora:
Looks good to me

certora/specs/RelativeCaps.spec
```diff
@@ -5,35 +5,34 @@ using Utils as Utils;
...
-
-    bytes32[] ids;
-    uint i;
-    havoc currentContract.caps[ids[i]].allocation assuming i < ids.length;
-
-    return ids;
-}
-
-// Check that allocation are within relative caps limit assuming no deallocation and no relative cap decrease.
+// Check that no function can break the cap limit, except for:
```
adhusson, line 11:
```suggestion
// Check that no function can break the cap limit, except:
```

certora/specs/RelativeCaps.spec
```diff
@@ -5,35 +5,34 @@ using Utils as Utils;
...
+         f.selector != sig:decreaseRelativeCap(bytes, uint256).selector &&
+         f.selector != sig:deallocate(address, bytes, uint256).selector
+} {
     bytes32 id;
+    // Tracks the firstTotalAssets value after calling f.
+    uint256 firstTotalAssetsAfter;
 
     require currentContract.caps[id].relativeCap < Utils.wad() =>
-    currentContract.caps[id].allocation <= (currentContract.firstTotalAssets * currentContract.caps[id].relativeCap) / Utils.wad();
+    currentContract.caps[id].allocation <= (firstTotalAssetsAfter * currentContract.caps[id].relativeCap) / Utils.wad();
```
adhusson, line 29: I don't understand the point of the first require since the value of `firstTotalAssetsAfter` is constrained after the call anyway. It prevents the overflow of `id.allocation * wad / id.relativeCaps` but I don't know what the point is.

adhusson (approved):

certora/specs/Invariants.spec
```diff
@@ -67,5 +67,21 @@ strong invariant decreaseTimelockTimelock()
 strong invariant totalSupplyIsSumOfBalances()
     totalSupply() == sumOfBalances;
 
+strong invariant registeredAdaptersAreSet()
+    (forall uint256 i. i < currentContract.adapters.length => currentContract.isAdapter[currentContract.adapters[i]])
```
MathisGD, line 71: is there a reason to do an explicit forall, rather than verifying for an arbitrary i

QGarchery: Actually, thinking about it, I don't think it can be done easily. The reason is that to prove `adaptersUnique` we need the property that `isAdapter[adapters[i]]`, otherwise the function `addAdapter` might add a duplicate. But the index `i` is itself quantified (so we can't do `requireInvariant registeredAdaptersAreSet(i)` in the proof of `adaptersUnique`. 

We could try to also make `adaptersUnique` into `adaptersUnique(uint256 i, uint256 j)`, but then to prove `registeredAdaptersAreSet(k)` we would need to require indices that are difficult to know. More precisely, what we need to know for the proof of `adaptersUnique`, when the function `removeAdapter(account)` is called, is that there is no duplicate of `account`. In terms of spec this would mean calling `requireInvariant adaptersUnique(i, k)` where `i` is the index that was removed from the list. We could find this index with some munging but it's a lot of overhead for not much IMO

certora/specs/RelativeCaps.spec
```diff
@@ -0,0 +1,38 @@
...
+    // Tracks the firstTotalAssets value after calling f.
+    uint256 firstTotalAssetsAfter;
+
+    require currentContract.caps[id].relativeCap < Utils.wad() =>
+    currentContract.caps[id].allocation <= (firstTotalAssetsAfter * currentContract.caps[id].relativeCap) / Utils.wad();
+
+    f(e, args);
+
+    // Note that firstTotalAssets is not reset after f, because in CVL functions calls are not isolated in different transactions.
+    require firstTotalAssetsAfter == currentContract.firstTotalAssets;
```
MathisGD, line 34: the thing we want to avoid is the bad debt realization right? so why don't we explicitly check that (totalAssets decreasing), or do something like in ExchangeRate.spec

PR #705, opened 2025-08-20
[Certora] relative caps are not violated invariant

colin-morpho:
This PR is an alternative to #704:
  - check relative caps are not violated using an invariant assuming no deallocation or that the relative cap is decreased;
  - check registered adapters are unique.


QGarchery:
Closing because it isn't modeling the situation well: for `firstTotalAssets == 0` the requirement on the allocation is too strong (it requires that the allocation is 0)

PR #712, opened 2025-08-26
document position underestimate

adhusson:
fixes https://github.com/sherlock-audit/2025-08-morpho-vault-v2-aug-13th/issues/36

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -68,8 +68,11 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
-/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance due to roundings or entry/exit fees.
-/// @dev This underestimate is only resolved in the next transaction, and so it should be either always negligible to avoid excessive fees and withdrawals at unfair share prices, or it should be expensive to inflate, for instance by iterating deposits.
+/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance
+/// due to roundings or entry/exit fees.
+/// @dev This underestimate is only resolved in the next transaction, and so it should be either always negligible to
+/// avoid excessive fees and withdrawals at unfair share prices, or it should be expensive to inflate, for instance by
+/// iterating deposits.
 ///
 /// LIQUIDITY ADAPTER
 /// @dev Liquidity is allocated to the liquidityAdapter on deposit/mint, and deallocated from the liquidityAdapter on
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -68,6 +68,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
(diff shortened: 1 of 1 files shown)


src/VaultV2.sol
```diff
@@ -68,6 +68,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
+/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance due to roundings or entry/exit fees.
+/// @dev This understimate is only resolved in the next transaction, and so it should be small to avoid excessive fees and withdrawals at unfair share prices.
```
MathisGD, line 72: should we say that the adapter might not statically prevent it, but that it should be used in ways that prevent it (like for re-entrance)?

MathisGD, line 72: the main problem is the fact that allocators can loop allocate/deallocate, so it should stay small notably compared to the gas (the comment right now don't make me think about that)

adhusson: I don't find a good phrasing for this, do you have an idea? The issue is that it addresses possible future adapters (some might prevent it statically, some might not).

src/VaultV2.sol
```diff
@@ -68,6 +68,11 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
+/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance
+/// due to roundings or entry/exit fees.
+/// @dev This underestimate is only resolved in the next transaction, and so it should be either always negligible to
+/// avoid excessive fees and withdrawals at unfair share prices, or it should be expensive to inflate, for instance by
+/// iterating deposits.
```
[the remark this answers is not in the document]
adhusson:
```suggestion
/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance
/// due to roundings or entry/exit fees. This underestimate should stay negligible compared to gas. Adapters might not statically ensure this, but the curators should not interact with markets that can create big entry/exit losses.
```

ok for me (fixed an extra word)

src/VaultV2.sol
```diff
@@ -68,6 +68,9 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
+/// @dev On allocation or deallocation, the underlying position of an adapter may become underestimated, for instance
```
MathisGD, line 71:
```suggestion
/// @dev On allocation or deallocation, the adapter might loose some assets, for instance
```

MathisGD (approved):

src/VaultV2.sol
```diff
@@ -72,6 +72,9 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
+/// @dev On allocate or deallocate, the adapters might loose some assets (total realAssets decreases), for instance due
```
adhusson, line 75:
```suggestion
/// @dev On allocate or deallocate, the adapters might lose some assets (total realAssets decreases), for instance due
```

MathisGD (approved):

peyha (approved):

PR #713, opened 2025-08-26
fix: document remove adapter scenario

MathisGD:
fixes https://github.com/sherlock-audit/2025-08-morpho-vault-v2-aug-13th/issues/33


src/VaultV2.sol
```diff
@@ -68,6 +68,9 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// the same ids.
 /// @dev If allocations underestimate the actual assets, some assets might be lost because deallocating is impossible if
 /// the allocation is zero.
+/// @dev Except particular scenarios, adapters should be removed only if they have no assets. In order to ensure no
+/// allocator can allocate some assets in an adapter being removed, it should have an adapter id and its caps should be
```
adhusson, line 72:
```suggestion
/// allocator can allocate some assets in an adapter being removed, there should be an id exclusive to the adapter with its cap set to zero.
```

PR #716, opened 2025-08-26
fix: document caps can be exceeded

MathisGD:
fixes https://github.com/sherlock-audit/2025-08-morpho-vault-v2-aug-13th/issues/39

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -44,7 +44,7 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The allocation is not always up to date, because interest are added only when (de)allocating in the
 /// corresponding markets, and losses are deducted only when realized for these markets.
 /// @dev The caps are checked on allocate (where allocations can increase) for the ids returned by the adapter.
-/// @dev Relative caps are "soft" in the sense that they are only checked on allocate.
+/// @dev Relative caps are "soft" in the sense that they not checked on exit.
 /// @dev Caps can be exceeded because of interest.
 /// @dev The relative cap is relative to totalAssets, or more precisely to firstTotalAssets.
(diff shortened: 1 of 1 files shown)


src/VaultV2.sol
```diff
@@ -45,6 +45,7 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// corresponding markets, and losses are deducted only when realized for these markets.
 /// @dev The caps are checked on allocate (where allocations can increase) for the ids returned by the adapter.
 /// @dev Relative caps are "soft" in the sense that they are only checked on allocate.
```
adhusson, line 47:
```suggestion
/// @dev Relative caps are "soft" in the sense that they not checked on exit.
```

adhusson (approved):

src/VaultV2.sol
MathisGD: no it's the fact that they are not checked on withdraw!

MathisGD: @adhusson why are you proposing this?

src/VaultV2.sol
```diff
@@ -44,7 +44,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// @dev The allocation is not always up to date, because interest are added only when (de)allocating in the
 /// corresponding markets, and losses are deducted only when realized for these markets.
 /// @dev The caps are checked on allocate (where allocations can increase) for the ids returned by the adapter.
-/// @dev Relative caps are "soft" in the sense that they are only checked on allocate.
+/// @dev Relative caps are "soft" in the sense that they not checked on exit.
```
peyha, line 47:
```suggestion
/// @dev Relative caps are "soft" in the sense that they are not checked on exit.
```

adhusson (approved):

PR #721, opened 2025-08-26
fix: document donation/penalty stealing

MathisGD:
fixes https://github.com/sherlock-audit/2025-08-morpho-vault-v2-aug-13th/issues/43

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -34,6 +34,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// share price but not faster than the maxRate, and the interest are accrued only once per transaction. In order to
 /// protect against inflation attacks, the vault might need to be seeded with an initial deposit. See
 /// https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
+/// @dev Donations and forceDeallocate penalties can be in part stolen by opportunistic depositors if the maxRate is
+/// much greater than the rate.
 ///
 /// CAPS
 /// @dev Ids have an asset allocation, and can be absolutely capped and/or relatively capped.


adhusson (approved):

src/VaultV2.sol
```diff
@@ -34,6 +34,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// share price but not faster than the maxRate, and the interest are accrued only once per transaction. In order to
 /// protect against inflation attacks, the vault might need to be seeded with an initial deposit. See
 /// https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
+/// @dev Donations and forceDeallocate penalties can be in part stolen by opportunistic depositors if the maxRate is
+/// much greater than the rate.
```
peyha, line 38: isn't it the opposite ? rate greater than maxRate

MathisGD: let's make an example: the rate averages at around ~5%. The maxRate is 200% (the maxMaxRate). Someone makes a forceDeallocate, with a penalty. Let's say that this penalty makes it such that during 10 blocks, the vault will receive 200%. An opportunistic depositor could come in just for these 10 blocks (and go from vault to vault like that). Whereas if you keep the maxRate close to the rate (let's say 10%), you reduce by a lot the incentive to do that.

Note that this makes sense w.r.t. the penalty. Its goal is to compensate other depositors for the potential loss of yield before the rebalancing, not to make them win more.

About if it is fine for the allocators to have this role (of setting the maxRate), it is because they can already allocate the vault in such a way that it doesn't earn anything anymore (they are trusted for the future yield).

MathisGD: right, it's misleading (because it's actually not a problem that the earned rate is way above the maxRate). I'll change with your prop.

src/VaultV2.sol
```diff
@@ -34,6 +34,8 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// share price but not faster than the maxRate, and the interest are accrued only once per transaction. In order to
 /// protect against inflation attacks, the vault might need to be seeded with an initial deposit. See
 /// https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
+/// @dev Donations and forceDeallocate penalties can be in part stolen by opportunistic depositors if they make the rate
+/// increase by a large factor. The maxRate can be used to prevent that.
```
peyha, line 38:
```suggestion
/// increase by a large factor. Setting a sufficiently high maxRate prevents that.
```
feel like it's a bit more accurate

peyha (approved):
After thinking about it, this fix is better than handling `penaltyAssets` separately as suggested in the audit because
- vault managers will already set the max rate to handle spikes of interests (possible with market V2)
- it's not so much in practice, a 1% force deallocate with a 1% fee gives 1 bps of additional assets which corresponds to 3.65% extra rate if its distributed over 24h
- it does not fix the issue for donation (even though donations are way less likely than forceDeallocate)

src/VaultV2.sol
```diff
@@ -34,6 +34,9 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// share price but not faster than the maxRate, and the interest are accrued only once per transaction. In order to
 /// protect against inflation attacks, the vault might need to be seeded with an initial deposit. See
 /// https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
+/// @dev Donations and forceDeallocate penalties can be in part stolen by opportunistic depositors if they make the rate
```
adhusson, line 37:
```suggestion
/// @dev If they make the rate increase by a large factor, donations and forceDeallocate penalties can be in part stolen
```

src/VaultV2.sol
```diff
@@ -34,6 +34,9 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
...
+/// increase by a large factor. Setting a low maxRate prevents that by making the donation/penalty distributed over a
```
adhusson, line 38:
```suggestion
/// by opportunistic depositors. Setting a low maxRate prevents that by making the donation/penalty distributed over a
```

src/VaultV2.sol
```diff
@@ -34,6 +34,9 @@ import {ISharesGate, IReceiveAssetsGate, ISendAssetsGate} from "./interfaces/IGa
 /// share price but not faster than the maxRate, and the interest are accrued only once per transaction. In order to
 /// protect against inflation attacks, the vault might need to be seeded with an initial deposit. See
 /// https://docs.openzeppelin.com/contracts/5.x/erc4626#inflation-attack
+/// @dev If they make the rate increase by a large factor, donations and forceDeallocate penalties can be in part stolen
```
QGarchery, line 37: I don't like the word "stolen" here, because nothing says that those funds should go to the current depositors (and not future depositors, even if we could argue that they are "opportunistic")

PR #727, opened 2025-08-27
feat: adapter whitelist

MathisGD:
TODO

### DIFF
--- a/test/SettersTest.sol
+++ b/test/SettersTest.sol
@@ -219,7 +219,7 @@ contract SettersTest is BaseTest {
         assertFalse(vault.isAllocator(newAllocator));
     }
 
-    function testAddAdapter(address rdm) public {
+    function testAddAdapterNoWhitelister(address rdm) public {
         vm.assume(rdm != curator);
         address newAdapter = makeAddr("newAdapter");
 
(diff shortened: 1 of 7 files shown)


qdrn:
One "weird" thing is that you can add adapters and then a whitelister and end up in a state where none of the adapters are whitelisted...

MattLsb:
>  If the old adapter is removed, existing vaults can't use the new adapter without removing the old adapter (which is extremely heavy)
I’m not sure I understand. There’s no requirement to remove the old adapter. The issue only comes up if a new registry is added without including the old adapter, not when the existing registry is updated.

If there’s no critical issue with the old adapter, it’s fine for the registry to list both the old and new adapters.

That said, if an old adapter does have a serious problem, there’s currently no way to enforce aligned vaults to remove it via policy (and none of the proposed options solve this).

MattLsb:
I agree. Note that this makes checking whether a vault is ok a bit more complex.

For example in the app:

check if the registry is Morpho and addRegister is abdicated
vs
check the vault's adapters and compare them against the Morpho registry (past and present).

Still seems fine overall though.

MathisGD:
convinced that we should leave the behavior as is. The main argument is that constraining future adapters is useful for businesses (like distributors or morpho itself) but not users, so it's fine to have a bit more complexity in the use (to check the "alignment").

peyha:
The Readme could be updated in the adapter/curator part (I feel like we stopped updating it recently?)

src/interfaces/IAdapterRegistry.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2025 Morpho Association
+pragma solidity >=0.5.0;
+
+interface IAdapterRegistry {
+    function isValidAdapter(address account) external view returns (bool);
```
MathisGD, line 6: Wdyt?
```suggestion
    function isInRegistry(address account) external view returns (bool);
```

adhusson: Will update

peyha (approved):
what was the reason for checking all current adapter in `setAdapterRegistry` in the end ?

adhusson:
> what was the reason for checking all current adapter in setAdapterRegistry in the end ?

So you have the property that if a vault uses an add-only adapter registry (a registry that never un-whitelists an adapter), that vault can only ever have adapters approved by the registry.

It means fewer things to check from the outside.

test/SettersTest.sol
```diff
@@ -239,6 +239,26 @@ contract SettersTest is BaseTest {
         assertEq(vault.adapters(0), newAdapter);
     }
 
+    function testAddAdapterRegistry(bool isInRegistry) public {
```
peyha, line 242: a bit weird that this bool has the same name as the function in `IAdapterRegistry` but no big deal

peyha (approved):
~~we could add a test when an adapter is set in the registry and then unset to check that vaults revert but it's an edge case so not mandatory imo~~

PR #731, opened 2025-08-27
docs: fix min time to function execution

adhusson:
To execute any function, it is also possible to decrease its timelock to zero, and then execute it.


src/VaultV2.sol
MathisGD, line 1: timelock[selector] <= timelock[decreaseTimelock], so it doesn't change anything right?

PR #733, opened 2025-08-28
fix: document removal from market list

MathisGD:
fixes https://cantina.xyz/code/36c2cec0-15bb-4aaa-8ad9-93e08fa71081/findings/4


adhusson:
Although this could be a case of overcommenting

MathisGD:
why? it's less specific, it depends on the adapter etc

peyha:
but it could also happen to the vault v1 adapter no ?

MathisGD:
there is no concept of list of markets there

peyha (approved):

PR #735, opened 2025-08-28
fix: document fistTotalAssets edge case when the vault is empty

MathisGD:
fixes https://cantina.xyz/code/36c2cec0-15bb-4aaa-8ad9-93e08fa71081/findings/11


peyha (approved):

src/VaultV2.sol
```diff
@@ -57,6 +57,8 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 /// @dev This mechanism can generate false positives on relative cap breach when such a cap is nearly reached,
 /// for big deposits that go through the liquidity adapter.
 /// @dev Relative caps can still be manipulated by allocators (with short-term deposits), but it requires capital.
+/// @dev The behavior of firstTotalAssets is different when the vault has totalAssets=0, but it does not matter
+/// internally because in this case there are not investments to cap.
```
adhusson, line 61:
```suggestion
/// internally because in this case there are no investments to cap.
```

PR #722, opened 2025-08-26
automatic timelock duration for decrease timelock

adhusson:
with this change we could completely remove abdicate

PR #737, opened 2025-08-29
auto `decreaseTimelock` timelock, with 1 year cap

adhusson:
* the timelock to decrease a timelock for a selector is the timelock of that selector
* increaseTimelock is timelocked
* 1 year global cap on timelocks for sanity
* keep abdicate

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -318,7 +318,6 @@ contract VaultV2 is IVaultV2 {
         require(executableAt[data] == 0, ErrorsLib.DataAlreadyPending());
 
         bytes4 selector = bytes4(data);
-
         uint256 _timelock = timelock[selector == IVaultV2.decreaseTimelock.selector ? bytes4(data[4:8]) : selector];
 
         require(_timelock != type(uint256).max, ErrorsLib.Abdicated());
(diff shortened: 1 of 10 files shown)


src/VaultV2.sol
```diff
@@ -398,21 +397,33 @@ contract VaultV2 is IVaultV2 {
...
 
         timelock[selector] = newDuration;
         emit EventsLib.IncreaseTimelock(selector, newDuration);
     }
 
+    /// @dev Irreversibly disable submit for a selector.
+    /// @dev Be particularly careful as this action is not reversible.
+    /// @dev Existing timelocked operations submitted before abdicating the selector can still be executed. The
+    /// abdication of a selector only prevents future operations to be submitted.
+    function abdicateSubmit(bytes4 selector) external {
```
MathisGD, line 416: I wasn't thinking about separated function, but works too

adhusson: I started having an `increaseTimelock` that checks `<= 1 year || == uint.max` but it really called for a separate function. In code but also you can have a bigger timelock for abdicate.

peyha:
not a huge fan of this version, the 1 year limit is arbitrary. What if a vault wants to guarantee his users that he won't change the performance/management fee for the next 2 years ?

adhusson:
Closing in favor of #722, `increaseTimelock` should just be used carefully.

PR #748, opened 2025-09-10
audit: chainsecurity

MathisGD:


adhusson (commented):
The date on the file is 

Image

MathisGD:
the date I put is the date of the tag of the last audited commit

adhusson:
In that case it's commit 275c2f9cbe6d79e8e4712497dc3cb51cc70f786f which is from september 4th I think

adhusson:
for reference the tag is the day the tag was made, the audit is from the 05, the last commit is from 04, not an exact match but it works

peyha (approved):

PR #753, opened 2025-09-14
fix: document that firstTotalAsset is used to only accrue once

MathisGD:


adhusson (approved):

src/VaultV2.sol
```diff
@@ -58,6 +58,7 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 /// @dev The variable firstTotalAssets tracks the total assets after the first interest accrual of the transaction.
 /// @dev Used to implement a mechanism that prevents bypassing relative caps with flashloans. This mechanism makes the
 /// caps conservative and can generate false positives, notably for big deposits that go through the liquidity adapter.
+/// @dev Also used to accrue interest only once per transaction (except if the vault is empty, see "share price" above).
```
adhusson, line 61: actually the part in parenthesis is not clear at all

PR #758, opened 2025-09-15
fix: abdicated mapping

MathisGD:
we found a bug: timelock[selector]==max && pendingCount[selector]==0 doesn't mean that the function can't be called anymore, because there could well be a pending decreaseTimelock(selector, x). One solution was to count pendingDecreaseTimelock (see #757) or to prevent setting decreaseTimelock if the timelock is max (see #756), but this one was generally cleaner and simpler.


peyha (approved):

src/VaultV2.sol
```diff
@@ -461,6 +462,12 @@ contract VaultV2 is IVaultV2 {
         emit EventsLib.DecreaseTimelock(selector, newDuration);
     }
 
+    function abdicate(bytes4 selector) external {
+        timelocked();
+        abdicated[selector] = true;
+        emit EventsLib.Abdicate(selector);
+    }
```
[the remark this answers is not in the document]
Saw-mon-and-Natalie: The above regarding disabling the abdication itself (which applies globally for all timelocked endpoints) was not previously possible with the previous setup of setting really high time locks and making sure the pending counts for the endpoints and also its timelock decrease were `0`.

dmitriia (commented):
It became possible to front run execution of valid submits with new `abdicate()`, cancelling them, intentionally or not. That behavior can be described in the comments.

Also, `type(uint256).max` comment no longer looks relevant, L338-339 can be, e.g.:
```solidity
/// @dev Will revert if the timelock value overflows when added to the block timestamp.
```

xiaoming9090:
The fix/change looks good to me. The new approach is cleaner.

src/VaultV2.sol
```diff
@@ -342,26 +345,24 @@ contract VaultV2 is IVaultV2 {
...
-        pendingCount[selector]++;
-        emit EventsLib.Submit(selector, data, executableAt[data], pendingCount[selector]);
+        emit EventsLib.Submit(selector, data, executableAt[data]);
     }
 
     function timelocked() internal {
+        bytes4 selector = bytes4(msg.data);
         require(executableAt[msg.data] != 0, ErrorsLib.DataNotTimelocked());
         require(block.timestamp >= executableAt[msg.data], ErrorsLib.TimelockNotExpired());
+        require(!abdicated[selector], ErrorsLib.Abdicated());
```
beiraoCertora, line 355: A previously submitted transaction will not execute if the function is abdicated. In my opinion, moving this check, `require(!abdicated[selector], ErrorsLib.Abdicated())`, into the `submit()` function makes more sense.

For example, let's say the curator wants to freeze the management fee, but first he wants to change its value.
One mistake could be submitting these two transactions at the same time, but executing them in the wrong order.

You might want to add back the `pendingCount` mapping and check if there is any submitted tx left before abdicating.

It's not critical, but in general, I think the submission order should be the same as the execution order. But fixing this requires more significant changes.

beiraoCertora (commented):
The fix looks good. Waiting for your feedback on one minor issue.

src/VaultV2.sol
MathisGD: the fact that abdicated=true prevents the execution of previously submitted data (and data submitted later of course) is an important feature, and the reason behind this change: we want to be able to simply (and onchain) assess if a value won't change anymore. With what you propose, it would be impossible because you would need to know about all the pending data to assert that (which impossible onchain).

one thing we could change is preventing submission after the abdication (because they won't be executable anyway), but as there could already be some unexecutable pending data, we decided that we shouldn't (it's more consistent this way, only the execution is prevented).

beiraoCertora: This makes me think of a new scenario. I'm trying to challenge the system.

Let's say there is data that the curator forgot to execute or revoke.
Now, let's say the curator wants to abdicate the `setPerformanceFeeRecipient`. The curator sends the transaction to abdicate, but just before it is executed, someone front-runs the abdication with the past data (address) that the curator forgot.
Now, the fee recipient is incorrect, and `setPerformanceFeeRecipient` is abdicated.

MathisGD: Btw, all this is supposed to be well documented in the code's comment. If you spot something missing, we should add it!

PR #760, opened 2025-09-15
test abdicated mapping

adhusson:


MathisGD (commented):
why one test per function? we don't have this for the timelock neither

PR #767, opened 2025-09-23
docs: submit change to timelock of decreaseTimelock

QGarchery:
See https://github.com/morpho-org/vault-v2/pull/722#discussion_r2368526841


src/VaultV2.sol
```diff
@@ -127,6 +127,8 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 /// TIMELOCKS
 /// @dev The timelock duration of decreaseTimelock is the timelock duration of the function whose timelock is being
 /// decreased (e.g. the timelock of decreaseTimelock(addAdapter, ...) is timelock[addAdapter]).
+/// @dev It is still possible to submit a change of the timelock duration of decreaseTimelock, but it won't have any
+/// effect and trying to execute this change will revert.
```
MathisGD, line 131:
```suggestion
/// @dev It is still possible to submit changes of the timelock duration of decreaseTimelock, but it won't have any
/// effect (and trying to execute this change will revert).
```

PR #769, opened 2025-09-24
Reentrancy view

QGarchery:


certora/specs/ReentrancyView.spec
```diff
@@ -0,0 +1,68 @@
...
+
+hook ALL_SSTORE(uint _, uint _) {
+    storageChanged = true;
+    if (staticCallAfterSStore) {
+        staticCallUnsafe = true;
+    }
+}
+
+hook STATICCALL(uint256 g, address addr, uint256 argsOffset, uint256 argsLength, uint256 retOffset, uint256 retLength) uint256 rc {
+    if (!ignoredStaticcall && storageChanged && addr != 0x1 && addr != currentContract) {
```
adhusson, line 48: Why != 0x1 and why exclude self-static call?

certora/specs/ReentrancyView.spec
```diff
@@ -0,0 +1,68 @@
...
+    ignoredStaticcall = false;
+}
+
+// Check that there are no reentrancy unsafe calls except potentially for balanceOf on the asset, realAssets on the adapters and canReceiveShares, canSendShares, canReceiveAssets and canSendAssets on the gates.
+rule reentrancyViewSafe(method f, env e, calldataarg data)
+filtered {
+    // forceDeallocate is a composition of deallocate and withdraw.
+    f -> f.selector != sig:forceDeallocate(address, bytes, uint256, address).selector
+} {
+    require !ignoredStaticcall == false, "setup ghost state";
```
adhusson, line 60: Why not `ignoredStaticcall == true`.


Also `ignoredStaticcall` is equivalent to "there was no static call or there was a staticcall but it was ignored"? It could be named `noUnsafeStaticcall`. Or even flipped and named `unsafeStaticcall`.

QGarchery: Whoops that's a mistake, it should be `ignoredStaticcall == false`. It actually catches more cases when I fix it.

This boolean is saying "the current STATICCALL should not be considered for violations", it's useful to exclude false positives notably. So for example if you have `accrueInterest(); otherContract.viewCall(); ... some SSTORE` then you know that this `otherContract.viewCall()` is view reentrancy safe, so you mark it as ignored

certora/specs/ReentrancyView.spec
QGarchery: 0x1 is ecrecover, it's just doing a pure computation so we can safely ignore it. For the self-static call I think it's an artifact of the Reentrancy.spec, but I think we can remove it

certora/specs/ReentrancyView.spec
QGarchery: Fixed now, I also added this new case in the notion page

certora/specs/ReentrancyView.spec
QGarchery: I added a comment for 0x1, and I removed the exclusion of self-static calls

PR #780, opened 2025-10-16
[Certora] Review chainsec specs

QGarchery:


certora/specs/AllocationVaultV2.spec
```diff
@@ -6,36 +6,46 @@ methods {
     function allocation(bytes32) external returns uint256 envfree;
 }
 
-// Check that the only functions able to change allocations are deposit, mint, withdraw, redeem, allocation, deallocate, forceDeallocate.
+//E RUN : https://prover.certora.com/output/7508195/e366be539f21406aac7fded3fbf53a5c/?anonymousKey=bfb434d7be5020b47cc637f24dcbc70f767c8138
```
QGarchery: I'm guessing this is useful to get back the previous working run once it gets broken. But since we keep developing until its ready (and then it is immutable), it makes less sense in our case I think. 
Also curious what the `E RUN`, `RUN` and `RULE` syntax mean in those cases

certora/specs/AllocationVaultV2.spec
```diff
@@ -6,36 +6,46 @@ methods {
...
-         f.selector != sig:forceDeallocate(address,bytes,uint,address).selector
+    f -> !f.isView && !canChangeAllocation(f)
 } {
     bytes32 id;
     uint256 allocationPre = allocation(id);
 
     f(e, args);
 
-    assert allocation(id) == allocationPre;
+    assert allocation(id) == allocationPre, "Only authorized functions should change allocations";
```
QGarchery: I don't feel like it adds much in the case of `assert`s, because they are already documented at the top of the rule

certora/specs/AllocationVaultV2.spec
```diff
@@ -6,36 +6,46 @@ methods {
...
+    f -> isERC4626Function(f)
 } {
     bytes32 id;
     uint256 allocationPre = allocation(id);
 
     f(e, args);
 
-    assert allocation(id) != allocationPre => currentContract.liquidityAdapter != 0;
+    // If allocation changed, then liquidity adapter must be set
+    assert allocation(id) != allocationPre => currentContract.liquidityAdapter != 0, "Allocations should only change via liquidity adapter in ERC4626 functions";
```
QGarchery: same

certora/specs/AllocationVaultV2.spec
lilCertora: It was more for your to have access to a working run in a fast way , this has to be removed of course when it's merged

certora/specs/AllocationVaultV2.spec
```diff
@@ -6,36 +6,46 @@ methods {
...
+
+// Definition: Functions that are allowed to change allocations
+definition canChangeAllocation(method f) returns bool =
+    f.selector == sig:deposit(uint256,address).selector ||
+    f.selector == sig:mint(uint256,address).selector ||
+    f.selector == sig:withdraw(uint256,address,address).selector ||
+    f.selector == sig:redeem(uint256,address,address).selector ||
+    f.selector == sig:allocate(address,bytes,uint256).selector ||
+    f.selector == sig:deallocate(address,bytes,uint256).selector ||
+    f.selector == sig:forceDeallocate(address,bytes,uint256,address).selector;
```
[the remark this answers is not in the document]
lilCertora: ok i understand, we will inline every definition used only once

PR #782, opened 2025-10-17
[Certora] Remove munging idea

QGarchery:


lilCertora:
if the idea is to remove munging it can be good to just do that no ?

```
// Check that loss realization decreases the share price.
rule lossRealizationDecreasesSharePrice(env e, address adapter, bytes data){
    mathint assetsBefore = assets();
    mathint sharesBefore = shares();
    
    // Capture _totalAssets before interest accrual
    uint128 totalAssetsBefore = currentContract._totalAssets;

    accrueInterest(e);
    
    // Capture _totalAssets after interest accrual
    uint128 totalAssetsAfter = currentContract._totalAssets;

    // Loss realization: new total assets is less than old total assets
    require (totalAssetsAfter < totalAssetsBefore, "assume loss realization");

    assert assets() * sharesBefore <= assetsBefore * shares();
}
```
[At least the rule hold](https://prover.certora.com/output/7508195/facb9d90ddad43f7b0f085308b2c24b8/?anonymousKey=57824fa2de5c91eab15137835eae7541b1925f4e) even if it's not perfect in the case where you add fees or something like that

PR #783, opened 2025-10-31
[Certora] Remove munging of VaultV2

QGarchery:
Idea is from @lilCertora


jhoenicke (approved):
Looks good to me.  Getting rid of that munge makes it much nicer and the check is still equivalent.

Do you have a rule for the inverse, i.e. no loss realization implies accrueInterest only increases share value?  Does it hold?

PR #786, opened 2025-11-07
[midnight adapter] Adapter

adhusson:
(eager loss realization)


src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+
+    /* ACCOUNTING */
+
+    uint256 public _totalAssets;
+    uint48 public lastUpdate;
+    uint48 public firstMaturity;
+    uint128 public currentGrowth;
+    mapping(uint256 timestamp => Maturity) public _maturities;
+    mapping(bytes32 obligationId => ObligationPosition) public _positions;
+    /* CONSTRUCTOR */
```
peyha:
```suggestion

    /* CONSTRUCTOR */
```

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+        require(msg.sender == address(morphoV2), NotMorphoV2());
+        require(buyer == address(this), NotSelf());
+        bytes32 obligationId = _obligationId(obligation);
+        uint48 prevMaturity = abi.decode(data, (uint48));
+        require(prevMaturity < obligation.maturity, IncorrectHint());
+
+        accrueInterest();
+        if (obligation.maturity > block.timestamp) {
+            uint128 timeToMaturity = uint128(obligation.maturity - block.timestamp);
+            uint128 gainedGrowth = ((obligationUnits - buyerAssets) / timeToMaturity).toUint128();
```
peyha: there could be some roundings by defining growth with this, not sure it's an issue though

peyha (commented):
Nice 🔥 🔥 using a linked list to manage maturities is smart

If I understand correctly, the allocation workflow should be

An allocator creates a make offer for the vault
A taker takes this offer onchain
| morphoV2 calls `onBuy` on the adapter
| | the adapter updates its state
| | the adapter calls allocate on its parent vault
| | | the vault V2 calls allocate on the market V2 adapter
| | | | the adapter returns the corresponding ids and allocated assets

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,451 @@
...
+        require(signer != address(0), IncorrectSigner());
+        require(IVaultV2(parentVault).isAllocator(signer), IncorrectSigner());
+
+        return CALLBACK_SUCCESS;
+    }
+
+    /// @dev `data` is used for new maturity insertions.
+    /// @dev It should encode a maturity present in the linked list.
+    /// @dev That maturity should be earlier than the inserted obligation maturity.
+    function onBuy(
```
peyha, line 282: would it be worth implementing a, additional callback in this function ? rationale is for rolling you might want to withdraw from the previous obligation in this call so the vault could do it in the callback. Two ways to achieve this
- implement an arbitrary callback, essentially converting `bytes memory data` to `(address callback, bytes memory callbackData)`
- implement a way to withdraw from one or more obligations, converting `bytes memory data` to `(bytes32, amount)[]` specifying a list of obligation ids and amount to withdraw

adhusson: - If the buy is initiated by the allocator, doing `withdrawToVault` then `buy` works. No need for callbacks.
- If the buy is initiated by the taker and there is a callback, the taker needs to know where to withdraw from. 
- If the buy is initiated by the taker and there is no callback, the taker cannot `withdrawToVault` to get the liquidity, but we can change that: we can make the `withdrawToVault` function permissionless.
- Yes arbitrary callbacks are a little tricky, in particular because if that sub-callback updates the position of the obligation currently being bought, that can break the adapter's internal accounting.
- It's true that doing a flashloan to work around this is not very direct:
```
- Allocator calls `vault.setLiquidityAdapterAndData(address(0), "")`, gets vault shares.
- Allocator flashloans `a`:
  - Allocator deposits flashloaned assets into idle.
  - Allocator calls `Midnight.take`:
    - Aadapter receives `a` assets from vault (in `adapter.onBuy`).
    - Midnight receives `a` assets from adapter.
  - Allocator calls `adapter.withdrawToVault(oldObligation, a)`, vault receives `a` assets.
  - Allocator redeems his vault shares.
- Allocator calls `vault.setLiquidityAdapterAndData(address(adapterBefore), dataBefore)`
```

peyha: >If the buy is initiated by the taker and there is no callback, the taker cannot withdrawToVault to get the liquidity, but we can change that: we can make the withdrawToVault function permissionless.

not very natural for the taker no ? he would have to use a multicall to `withdrawToVault` before taking (especially if he already has a callback)

>Yes arbitrary callbacks are a little tricky, in particular because if that sub-callback updates the position of the obligation currently being bought, that can break the adapter's internal accounting.

indeed, allocators could easily make mistakes and break the accounting so it would make sense to have very scoped callback (withdrawing from an other obligation, selling ?)

adhusson: > not very natural for the taker no ? he would have to use a multicall to withdrawToVault before taking (especially if he already has a callback)

True but anyway he will need to do some work? I don't imagine the maker picking the withdraw list in advance. Let's discuss live maybe.

adhusson: Thinking more about this I think the simplest is to
- not have a nested callback
- allow the midnight adapter to withdraw from a liquidity adapter

The allocators can separately do the work of withdrawing obligations to liquidity adapter or to idle in the vault.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,436 @@
...
+                zeroedDurationsIds[i] = keccak256(abi.encode("duration", packedDurations.get(newDurationCount + i)));
+            }
+            int256 change = -int256(uint256(maturityData.netCredit));
+            IVaultV2(parentVault).deallocate(address(this), abi.encode(zeroedDurationsIds, change), 0);
+        }
+    }
+
+    /* ACCRUAL */
+
+    function accrueInterestView() public view returns (uint48, uint128, uint256, uint256) {
```
adhusson: We could also do this, slightly shorter but harder to read (sum each crossed maturity's growth impact instead of summing total growth between each maturity):

```solidity
    function accrueInterestView() public view returns (uint48, uint128, uint256, uint256) {
      uint48 _firstMaturity = _maturities[0].nextMaturity;
      uint128 newGrowth = currentGrowth;
      uint256 newAvailableMaturities = availableMaturities;
      uint256 gainedAssets = 0;

      while (_firstMaturity != 0 && _firstMaturity < block.timestamp) {
          gainedAssets += uint256(_maturities[_firstMaturity].growth) * (_firstMaturity - lastUpdate);
          newGrowth -= _maturities[_firstMaturity].growth;
          _firstMaturity = _maturities[_firstMaturity].nextMaturity;
          newAvailableMaturities++;
      }

      gainedAssets += uint256(newGrowth) * (block.timestamp - lastUpdate);

      return (_firstMaturity, newGrowth, _totalAssets + gainedAssets, newAvailableMaturities);
  }
```

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,443 @@
...
+            maturityData.growth += growthIncrease;
+            currentGrowth += growthIncrease;
+        } else {
+            totalAssets += boughtNetCredit.toUint128();
+        }
+
+        maturityData.netCredit += boughtNetCredit.toUint128();
+        marketData.netCredit += boughtNetCredit.toUint128();
+
+        // Insert the maturity in the list if needed
```
peyha, line 326: would a dedicated lib for this kind of logic make sense ?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,443 @@
...
+        updateDurationCaps(market);
+
+        MaturityData storage maturityData = _maturities[market.maturity];
+        MarketData storage marketData = _markets[marketId];
+        uint256 timeToMaturity = market.maturity.zeroFloorSub(block.timestamp);
+        // current net credit cannot be > accounted net credit + bought net credit
+        uint256 netCreditLoss = uint256(marketData.netCredit) + boughtNetCredit - currentNetCredit(marketId);
+        decreaseNetCredit(marketId, market.maturity, netCreditLoss);
+
+        IVaultV2(parentVault)
```
peyha, line 305: the vault v2 doc specifies [here](https://github.com/morpho-org/vault-v2/blob/a02966caca748b1b69c86fd3825b99ece8c5a800/src/VaultV2.sol#L78) that "[adapters] must not re-enter (directly or indirectly) the vault" and this line is breaking this property, a few possibilities to move forward
1. Adapt the vault v2 doc to take this specific case into account
2. Use a new intermediary contract "Relayer" to handle the callback `onBuy`/`onSell` before calling the vault (but it adds operational complexity)
3. Refactor this function to make it clear that `onBuy` is not interacting dangerously with the vault

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+            _totalAssets += buyerAssets + (obligationUnits - buyerAssets) % timeToMaturity;
+            _positions[obligationId].growth += gainedGrowth;
+            _maturities[obligation.maturity].growthLostAtMaturity += gainedGrowth;
+            currentGrowth += gainedGrowth;
+        } else {
```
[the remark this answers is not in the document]
peyha: resolved as outdated

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+import {IERC20} from "../interfaces/IERC20.sol";
+import {SafeERC20Lib} from "../libraries/SafeERC20Lib.sol";
+import {MathLib} from "../libraries/MathLib.sol";
+import {MathLib as MorphoV2MathLib} from "lib/morpho-v2/src/libraries/MathLib.sol";
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IMorphoMarketV2Adapter, ObligationPosition, Maturity, IAdapter} from "./interfaces/IMorphoMarketV2Adapter.sol";
+
+/// @dev Approximates held assets by linearly accounting for interest separately for each obligation.
+/// @dev Losses are immdiately accounted minus a discount applied to the remaining interest to be earned, in proportion
+/// to the relative sizes of the loss and the adapter's position in the obligation hit by the loss.
```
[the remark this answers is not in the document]
peyha: resolved as outdated

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+    /* MANAGEMENT */
+
+    address public skimRecipient;
+    uint256 public minTimeToMaturity;
+    uint256 public minRate;
+
+    /* ACCOUNTING */
+
+    uint256 public _totalAssets;
+    uint48 public lastUpdate;
```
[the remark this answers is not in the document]
peyha: resolved as outdated

src/adapters/MorphoMarketV2Adapter.sol
peyha: resolved as outdated

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+        uint256 vaultRealAssets = IERC20(asset).balanceOf(address(parentVault));
+        uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
+        for (uint256 i = 0; i < adaptersLength; i++) {
+            vaultRealAssets += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
+        }
+        uint256 vaultBuffer = vaultRealAssets.zeroFloorSub(IVaultV2(parentVault).totalAssets());
+
+        uint256 _totalAssetsBefore = _totalAssets;
+        removeUnits(obligation, obligationUnits);
+        require(vaultBuffer >= _totalAssetsBefore.zeroFloorSub(_totalAssets), BufferTooLow());
```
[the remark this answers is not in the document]
peyha: resolved as outdated

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,443 @@
...
+        emit UpdateDurationCaps(market.maturity, newDurationCount, maturityData.netCredit);
+        // VaultV2.deallocate requires allocation > 0 for each returned id.
+        if (newDurationCount < oldDurationCount && maturityData.netCredit > 0) {
+            bytes32[] memory zeroedDurationsIds = new bytes32[](oldDurationCount - newDurationCount);
+            for (uint256 i = 0; i < zeroedDurationsIds.length; i++) {
+                zeroedDurationsIds[i] = keccak256(abi.encode("duration", packedDurations.get(newDurationCount + i)));
+            }
+            IVaultV2(parentVault)
+                .deallocate(address(this), abi.encode(zeroedDurationsIds, -int256(uint256(maturityData.netCredit))), 0);
+        }
```
peyha, line 149: another way to update these caps would be to return these ids/change in allocate/deallocate to let the vault update it automatically

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,447 @@
...
+            _maturities[maturityData.prevMaturity].nextMaturity = maturityData.nextMaturity;
+            _maturities[maturityData.nextMaturity].prevMaturity = maturityData.prevMaturity;
+            emit RemoveMaturity(maturity);
+        }
+    }
+
+    /// @dev Returns the number of durations in packedDurations that are at most the time to maturity.
+    function durationCount(uint256 maturity) internal view returns (uint256 count) {
+        uint256 timeToMaturity = maturity.zeroFloorSub(block.timestamp);
+        while (count < durationsLength && timeToMaturity >= packedDurations.get(count)) count++;
```
peyha, line 420: could speed this up with fenwick trees or segment tree + lazy propagation. not worth it for small duration array though

src/adapters/MidnightAdapter.sol
adhusson: I don't think we would gain anything in this case, unlike the DurationsLib's `get` it's not reused. Btw DurationLib's `pack` is not reused either in deployed code, it could be inlined, but it's useful to have it in tests.

src/adapters/MidnightAdapter.sol
adhusson: Good point. I think adapting the docs works in this case.

src/adapters/MidnightAdapter.sol
adhusson: I didn't get this.

src/adapters/MidnightAdapter.sol
adhusson: yes very likely too complex for 8 durations

src/adapters/MorphoMarketV2Adapter.sol
```diff
@@ -0,0 +1,321 @@
...
+        require(msg.sender == IVaultV2(parentVault).curator(), NotAuthorized());
+        require(_minTimeToMaturity <= type(uint48).max, IncorrectMinTimeToMaturity());
+        minTimeToMaturity = _minTimeToMaturity;
+    }
+
+    /* VAULT ALLOCATORS FUNCTIONS */
+
+    // Do not cleanup the linked list if we end up at 0 growth
+    function withdraw(Obligation memory obligation, uint256 obligationUnits, uint256 shares) external {
+        require(IVaultV2(parentVault).isAllocator(msg.sender), NotAuthorized());
```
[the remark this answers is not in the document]
adhusson: (note: it's been renamed `withdrawToVault` for clarity)

- It's safe to make it permissionless but I don't know if we want that
- Even if permissioned the user can buy at 1 through forceDeallocate then withdraw, it costs him the settlement fee + the penalty.

foundry.toml
```diff
@@ -3,13 +3,23 @@ via_ir = true
 optimizer = true
 optimizer_runs = 100000
 bytecode_hash = "none"
-evm_version = "cancun"
+evm_version = "osaka"
```
MathisGD: we can't change that for the whole repo

src/adapters/MidnightAdapterFactory.sol
```diff
@@ -0,0 +1,37 @@
...
+    mapping(address parentVault => mapping(address midnight => address)) public midnightAdapter;
+    mapping(address account => bool) public isMidnightAdapter;
+    uint256[] public durations;
+
+    /* CONSTRUCTOR */
+
+    /// @dev Durations are checked only when an adapter is created.
+    constructor(uint256[] memory _durations) {
+        durations = _durations;
+    }
```
MathisGD, line 20: why is it a factory constructor argument rather than per adapter?

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
```diff
@@ -0,0 +1,43 @@
...
+import {IMidnightAdapterSetterRatifier} from "./interfaces/IMidnightAdapterSetterRatifier.sol";
+
+/// @dev Sub-ratifier for MidnightAdapter offers: checks that the offer has been ratified onchain by an allocator of
+/// the maker adapter's parent vault in a Merkle tree of offers. To that end, it expects the ratifier data to contain
+/// the root of the tree, the leaf index of the offer in the tree, and the proof of the offer in the tree.
+/// @dev The root should correspond to the root of the offer tree, which is a Merkle tree of offers.
+/// @dev The leaf index determines each hash order during merkle proof verification.
+/// @dev A root remains ratified if the allocator that ratified it loses its role.
+/// @dev Same tree format as Midnight's SetterRatifier.
+contract MidnightAdapterSetterRatifier is IMidnightAdapterSetterRatifier {
```
MathisGD: wondering if we need the two ratifiers here. we could constrain allocators to use only one of the two methods (not fully sure which, but I tend towards ecrecover)

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
```diff
@@ -0,0 +1,43 @@
...
+        require(
+            IVaultV2(parentVault).isAllocator(msg.sender)
+                || (!newIsRootRatified && IVaultV2(parentVault).isSentinel(msg.sender)),
+            NotAuthorized()
+        );
+        isRootRatified[adapter][root] = newIsRootRatified;
+        emit SetIsRootRatified(msg.sender, adapter, root, newIsRootRatified);
+    }
+
+    function isRatified(Offer memory offer, bytes memory ratifierData, address) external view returns (bytes32) {
```
MathisGD: should we allow gated offers? seems like an important strategic decision

src/adapters/ratifiers/MidnightAdapterEcrecoverRatifier.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function isRatified(Offer memory offer, bytes memory ratifierData, address) external view returns (bytes32) {
+        (Signature memory sig, bytes32 root, uint256 leafIndex, bytes32[] memory proof) =
+            abi.decode(ratifierData, (Signature, bytes32, uint256, bytes32[]));
+        require(HashLib.isLeaf(root, HashLib.hashOffer(offer), leafIndex, proof), InvalidProof());
+        require(!isRootCanceled[offer.maker][root], RootCanceled());
+        bytes32 structHash = keccak256(abi.encode(HashLib.offerTreeTypeHash(proof.length), root));
+        bytes32 domainSeparator = keccak256(abi.encode(EIP712_DOMAIN_TYPEHASH, block.chainid, address(this)));
+        bytes32 digest = keccak256(bytes.concat("\x19\x01", domainSeparator, structHash));
+        // forge-lint: disable-next-item(ecrecover) offer sizes & cancellation protects against reuse.
```
MathisGD:
```suggestion
        // forge-lint: disable-next-item(ecrecover) signature is meant to be reused.
```

src/adapters/ratifiers/MidnightAdapterEcrecoverRatifier.sol
```diff
@@ -0,0 +1,52 @@
...
+        address signer = ecrecover(digest, sig.v, sig.r, sig.s);
```
MathisGD:
```suggestion
        address recovered = ecrecover(digest, sig.v, sig.r, sig.s);
```

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
MathisGD, line 1: do vaults need rate based offers? (https://github.com/morpho-org/midnight/pull/1180)

src/adapters/MidnightAdapterFactory.sol
adhusson: iirc it's from a live chat a few months ago i think the idea was to decide on standardized durations. easy to change

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: If you mean remove modularity, is it ok to be locked long term / require a new adapter if the EOAs want to use something other than secp256k1? (like secp256r1 for passkeys)

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,565 @@
...
+        timelocked();
+        abdicated[selector] = true;
+        emit Abdicate(selector);
+    }
+
+    function setSkipBufferCheck(bool newSkipBufferCheck) external {
+        timelocked();
+        skipBufferCheck = newSkipBufferCheck;
+        emit SetSkipBufferCheck(newSkipBufferCheck);
+    }
```
MathisGD, line 217: this is a very fundamental parameter. shouldn't it be set at construction?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,565 @@
...
+
+    function setSkimRecipient(address newSkimRecipient) external {
+        timelocked();
+        skimRecipient = newSkimRecipient;
+        emit SetSkimRecipient(newSkimRecipient);
+    }
```
MathisGD, line 229: if we remove `setSkipBufferCheck` ([link](https://github.com/morpho-org/vault-v2/pull/786/changes#r3948521384)), we can remove this (and remove the whole timelock system). it's not so fundamental that it's timelocked

src/adapters/MidnightAdapter.sol
adhusson: All the other parameters that can make a vault custodial are behind a timelock, like exit gates. This one feels different?

Also a vault with the buffer check could temporarily disable it to be able to sell credits that have lost a lot of their value and then reenable it?

src/adapters/MidnightAdapter.sol
adhusson: Note the skim recipient change is not timelocked in MorphoVaultV1Adapter, but it is timelocked in MorphoMarketV1AdapterV2.

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: The use case and how it would work (just a mapping?) is not clear to me, do we already have ideas on how it should work? If not, modularity means the gating can be added later.

foundry.toml
adhusson: fixed. fyi that param change did not change the bytecodes because of the fixed solidity pragmas.

src/adapters/MidnightAdapter.sol
MathisGD: you could always re-deploy an adapter with an other parameter right?

MathisGD: >Also a vault with the buffer check could temporarily disable it to be able to sell credits that have lost a lot of their value and then reenable it?

indeed but not sure that this is very important

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
MathisGD: yes, I realized later that it's pretty easy to change ratifiers. probably fine to go without

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
MathisGD: no I meant having only one of the two ratifiers available now. ratifier modularity is great (simple and scoped)

src/adapters/MidnightAdapterFactory.sol
MathisGD: it would be standardized but offchain enforcement for listed vaults is enough imo

src/adapters/MidnightAdapter.sol
adhusson: > you could always re-deploy an adapter with an other parameter right?

yes but it prevents enabling the skip buffer check to get out

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: will remove setter for now

src/adapters/MidnightAdapterFactory.sol
adhusson: done here https://github.com/morpho-org/vault-v2/pull/988

it could be annoying to query with the hash of durations so i added a getter that takes the durations array directly

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: same conclusion as https://github.com/morpho-org/vault-v2/pull/786#discussion_r3949350670, ok to start without?

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: done here https://github.com/morpho-org/vault-v2/pull/786/commits/1636f411a9ffdd4999dac946b76dc732c8b61b75

src/adapters/MidnightAdapter.sol
MathisGD: in a broken market, selling at a loss after the maturity to get out?

adhusson: yes, see [testAbandonMarketWithRevertingOracle](https://github.com/morpho-org/vault-v2/pull/786/changes#diff-c40c7c3f44d4bc5e1aebc80733ad6660b7340c465599b02b5a7bfd74f6922777R1731)

MathisGD: but then when deactivating the vault can be fully rugged, it's not a per market thing. isn't that problematic?

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
MathisGD: one could argue rate based is more important than price based though...

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,570 @@
...
+
+    /* VAULT ALLOCATORS FUNCTIONS */
+
+    function withdrawToVault(Market memory market, uint256 withdrawnAssets) external {
+        bytes32 marketId = IdLib.toId(market);
+        require(
+            IVaultV2(parentVault).isAllocator(msg.sender) || IVaultV2(parentVault).isSentinel(msg.sender),
+            NotAuthorized()
+        );
```
MathisGD, line 249: couldn't this be permissionless?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    /* RATIFIERS */
+
+    /// @dev Sub-ratifiers can approve any offer of the adapter that passes the checks of isRatified, akin to
+    /// allocators signing offer trees.
+    function setIsSubRatifier(address subRatifier, bool newIsSubRatifier) external {
+        require(
+            IVaultV2(parentVault).isAllocator(msg.sender)
+                || (!newIsSubRatifier && IVaultV2(parentVault).isSentinel(msg.sender)),
```
MathisGD, line 122: why can a sentinel remove?

src/adapters/MidnightAdapter.sol
adhusson: it could, see discussion [here](https://github.com/morpho-org/vault-v2/pull/786#discussion_r3812523190)

src/adapters/MidnightAdapter.sol
adhusson: a bad ratifier can approve offers that shouldn't be taken, so it seemed to fit the sentinel derisking role

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    function isRatified(Offer memory offer, bytes memory data, address taker) external view returns (bytes32) {
+        // Collaterals will be checked through vault ids.
+        require(offer.market.loanToken == asset, LoanAssetMismatch());
+        require(offer.maker == address(this), IncorrectMaker());
+        require(offer.callback == address(this), IncorrectCallbackAddress());
+        // For buy offers, Midnight enforces receiverIfMakerIsSeller == address(0).
+        require(offer.buy || offer.receiverIfMakerIsSeller == address(this), IncorrectReceiver());
+        require(offer.buy || offer.reduceOnly, NoDebtCreation());
```
MathisGD, line 136: maybe cleaner, but technically not mandatory since the adapter cannot supply collat, and noDebtWithoutCollat.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,570 @@
...
+        return CALLBACK_SUCCESS;
+    }
+
+    /* INTERNAL FUNCTIONS */
+
+    function currentNetCredit(bytes32 marketId) internal view returns (uint256) {
+        return
+            IMidnight(midnight).credit(marketId, address(this))
+                - IMidnight(midnight).pendingFee(marketId, address(this));
+    }
```
MathisGD, line 517: should accrue the loss on midnight?

src/adapters/MidnightAdapter.sol
adhusson: It's in the callback so the adapter could temporarily have debt, and it isn't clear that you can't then close the debt with another buy, will check.

src/adapters/MidnightAdapter.sol
adhusson: Not supposed to be needed because the loss is already accrued every time this is called, but it's missing in onSell actually.

src/adapters/MorphoMarketV2Adapter.sol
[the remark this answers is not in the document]
MathisGD: why wouldn't we want that? I don't see the downside. it allows for example to batch withdraw+withdraw from the vault

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,571 @@
...
+        // forge-lint: disable-next-item(reentrancy-no-eth) withdraw does not call back.
+        IMidnight(midnight).withdraw(market, withdrawnAssets, address(this), address(this));
+        // current net credit cannot be > accounted net credit
+        uint256 netCreditDecrease = _markets[marketId].netCredit - currentNetCredit(marketId);
+
+        decreaseNetCredit(marketId, market.maturity, netCreditDecrease);
+
+        // forge-lint: disable-next-item(unsafe-typecast) netCreditDecrease <= type(uint128).max.
+        IVaultV2(parentVault)
+            .deallocate(address(this), abi.encode(ids(market), -int256(netCreditDecrease)), withdrawnAssets);
```
MathisGD, line 262: to avoid this call, withdraw could happen in the deallocate function of this adapter no?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,571 @@
...
+    }
+
+    function take(Offer memory offer, bytes memory ratifierData, uint256 units) external {
+        require(IVaultV2(parentVault).isAllocator(msg.sender), NotAuthorized());
+        require(offer.market.loanToken == asset, LoanAssetMismatch());
+        IMidnight(midnight)
+            .take(
+                offer, ratifierData, units, address(this), offer.buy ? address(this) : address(0), address(this), hex""
+            );
+    }
```
MathisGD, line 273: same this could happen through allocate? is it useful?

src/adapters/MidnightAdapter.sol
adhusson: yes it could be replaced by a third branch in deallocate, you like it more?

src/adapters/MidnightAdapter.sol
adhusson: note it only delayed loss realization anyway

src/adapters/MidnightAdapter.sol
adhusson: (although then it can't be permissionless)

src/adapters/MidnightAdapter.sol
adhusson: It could but it would be ugly because `allocate` would have to repeat part of what `onBuy` does, and deallocate would have to repeat part of what of what `onSell` does ([slack discussion on take](https://morpholabs.slack.com/archives/C08A40KKN9W/p1787150859124259?thread_ts=1787149743.780129&cid=C08A40KKN9W)). It seems simpler/clearer to just have `take`.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,572 @@
...
+                    newGrowth -= _maturities[maturity].growth;
+                }
+            }
+        }
+        newTotalAssets += newGrowth * (block.timestamp - lastUpdate);
+
+        return (newGrowth, newTotalAssets);
+    }
+
+    function accrueInterest() public returns (uint128, uint256) {
```
MathisGD, line 316: there is no incentive to be calling this right? but it's the only way the adapter (and thus the vault) can see that it incurred losses on the markets. 

is that ok? I remember that at some point we were speaking about giving an incentive to people who come to realize losses.

is the accounting ok? the fact that we keep growing while the loss isn't accounted is fine?

adhusson: `accrueInterest` doesn't see losses because it does not go through every market. The only way to make the adapter realize a loss on a market is a buy, a sell (can be a forceDeallocate), or a withdraw. Withdraw 0 can ping a market, but for now it's not permissionless.

> is the accounting ok?

The accounting is only delayed. After the adapter realizes a loss the growth is reduced and the current assets are reduced. If someone deposits after a loss is realized on Midnight, but before the accounting of the adapter realizes it, then he could unfairly lose (similar to a possible unfair loss if you deposit before a loss is realized on Blue, but here the adapter also needs to see the loss).

It's possible to add a reward without changing the adapter: make an external contract with funds that can ping markets through the adapter. If `withdrawToVault` is not permissionless the contract can be an allocator/sentinel.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,572 @@
...
+        require(offer.maker == address(this), IncorrectMaker());
+        require(offer.callback == address(this), IncorrectCallbackAddress());
+        // For buy offers, Midnight enforces receiverIfMakerIsSeller == address(0).
+        require(offer.buy || offer.receiverIfMakerIsSeller == address(this), IncorrectReceiver());
+        require(offer.buy || offer.reduceOnly, NoDebtCreation());
+
+        (address subRatifier, bytes memory subData) = abi.decode(data, (address, bytes));
+        require(isSubRatifier[subRatifier], SubRatifierUnauthorized());
+        return IRatifier(subRatifier).isRatified(offer, subData, taker);
+    }
```
peyha, line 141: not a huge fan of having the ratifier inside the adapter. 
Would there be a way to let the dedicated ratifier handle all these checks and just have a `setRatifier` in the adapter to authorize it on midnight ?
It would separate the logic more cleanly imo

src/adapters/MidnightAdapter.sol
adhusson: Note we can also track all markets and ping them all at each realAssets. About 8k per market + the adapter needs to be able to go through all its markets.

src/adapters/MidnightAdapter.sol
adhusson: It's possible but I'm not sure we gain anything. Those other ratifiers would all have to duplicate the logic above (before the last 3 lines).

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,572 @@
...
+        address seller,
+        address,
+        bytes memory
+    ) external returns (bytes32) {
+        require(msg.sender == midnight, NotMidnight());
+        require(seller == address(this), NotSelf());
+
+        accrueInterest();
+        // forge-lint: disable-next-item(reentrancy-no-eth) updatePosition does not call back.
+        IMidnight(midnight).updatePosition(market, address(this));
```
MathisGD, line 487: it's already updated no?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,572 @@
...
+        bytes32[] memory idsArray = new bytes32[](1 + market.collateralParams.length * 2 + durationsCount);
+
+        uint256 j;
+        idsArray[j++] = adapterId;
+        for (uint256 i = 0; i < market.collateralParams.length; i++) {
+            address collateralToken = market.collateralParams[i].token;
+            idsArray[j++] = keccak256(abi.encode("collateralToken", collateralToken));
+            idsArray[j++] = keccak256(
+                abi.encode(
+                    "collateral", collateralToken, market.collateralParams[i].oracle, market.collateralParams[i].lltv
```
MathisGD: there should be "this" in one of the ids I think

src/adapters/MidnightAdapter.sol
adhusson: done here https://github.com/morpho-org/vault-v2/pull/990

it increases risk quite a bit
- if the adapter callback is not called the accounting is broken (also no caps check)
- the adapter could become a net borrower
- there is no buffer check so the allocator can sell everything for 0
- a bad ratifier can authorize other accounts

also it increases code size

src/adapters/MidnightAdapter.sol
MathisGD: btw there should be the LIF in there as well, no?

src/adapters/MidnightAdapter.sol
adhusson: the onSell callback happens after the onBuy callback, so the buyer could realize a loss in its callback

src/adapters/MidnightAdapter.sol
adhusson: there is just above! `idsArray[j++] = adapterId`

adhusson: see `/// @dev Liquidation cursors are omitted from collateral ids.` ([here](https://github.com/morpho-org/vault-v2/pull/786/changes/BASE..5f45107e48cd903007b2c88aeec098180b117cf2#diff-d8ed7cfd927b24af16996b4f64032f6f793d6832d1b346b2a5207895badd0d70R549)) it seemed too fine-grained, will re-add it

MathisGD: rationale is liquidation cursor = 1 is pretty dangerous, so you don't want to let allocators go into that

src/adapters/MidnightAdapter.sol
peyha: indeed it's less nice than I expected. ratifier calls are separated from callback calls but now a bad ratifier can be very dangerous. The current version is fine then
It also do not seem to allow new usecases. Both versions would allow a dutch auction sell offer with a correct subratifier/ratifier

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+
+    mapping(bytes4 selector => uint256) public timelock;
+    mapping(bytes4 selector => bool) public abdicated;
+    mapping(bytes data => uint256) public executableAt;
+
+    /* MANAGEMENT */
+
+    address public skimRecipient;
+    /// @dev Remaining allowance for losses not covered by the vault's buffer, in asset units.
+    uint256 public skipBufferAllowance;
```
peyha: not a fan of the naming (why "skip" ?)
prop: lossAllowance but might be too scary for users

peyha: thinking a bit more, not sure about this parameter. In practice it would be pretty difficult to set an accurate value because it would be an expectation of the loss wrt to the non accrued interest while having a timelock. I fear that in practice, allocators would just set it to a very high value in case of emergecy. Another version would be to use a boolean `canHaveLoss` on specific markets

adhusson: The allocator could loop through the market, unless canHaveLoss on a market prevents buying more into it.

peyha: yeah it would prevent buying more into it, it would be more similar to `burnShares` (though the whole market allocation is not necessarily lost)

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+        emit WithdrawToVault(marketId, withdrawnAssets, netCreditDecrease);
+    }
+
+    function take(Offer memory offer, bytes memory ratifierData, uint256 units) external {
+        require(IVaultV2(parentVault).isAllocator(msg.sender), NotAuthorized());
+        require(offer.market.loanToken == asset, LoanAssetMismatch());
+        IMidnight(midnight)
+            .take(
+                offer, ratifierData, units, address(this), offer.buy ? address(this) : address(0), address(this), hex""
+            );
```
peyha, line 272: if we don't check that this take is not accruing the debt of the adapter (which would revert anyway) we probably don't need to check that the offer is `reduceOnly` in `onRatify` for consistency

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+        bytes32[] memory idsArray = new bytes32[](1 + market.collateralParams.length * 2 + durationsCount);
+
+        uint256 j;
+        idsArray[j++] = adapterId;
+        for (uint256 i = 0; i < market.collateralParams.length; i++) {
+            address collateralToken = market.collateralParams[i].token;
+            idsArray[j++] = keccak256(abi.encode("collateralToken", collateralToken));
+            idsArray[j++] = keccak256(
+                abi.encode(
+                    "collateral",
```
peyha:
```suggestion
                    "collateralParams",
```
to follow the midnight struct

src/adapters/ratifiers/MidnightAdapterEcrecoverRatifier.sol
```diff
@@ -0,0 +1,52 @@
...
+/// @dev Sub-ratifier for MidnightAdapter offers: checks that the offer has been signed by an allocator of the maker
+/// adapter's parent vault in a Merkle tree of offers. To that end, it expects the ratifier data to contain the
+/// signature, the root of the tree, the leaf index of the offer, and the proof of the offer in the tree.
+/// @dev The root should correspond to the root of the offer tree, which is a Merkle tree of offers.
+/// @dev The leaf index determines each sibling's left/right position.
+/// @dev Hashing offers as in EIP-712, which allows clear signing of the tree, credits to Seaport for this mechanism.
+/// @dev Same tree format and signing scheme as Midnight's EcrecoverRatifier.
+/// @dev If block.chainid changes (hard fork), the EIP-712 domain separator changes and previously signed offers are
+/// no longer valid.
+contract MidnightAdapterEcrecoverRatifier is IMidnightAdapterEcrecoverRatifier {
```
peyha, line 22: shouldn't this ratifier only work for buy offers ?
Suppose I want to sell with another ratifier (e.g. through a dutch auction), then I would potentially need to increase `skipBufferAllowance` but then an allocator could sell with this ratifier at a worst price.

peyha: other possibility, have a `skipBufferAllowance` for each subratifier (not sure it makes sense)

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+        accrueInterest();
+        // forge-lint: disable-next-item(reentrancy-no-eth) updatePosition does not call back.
+        IMidnight(midnight).updatePosition(market, address(this));
+
+        uint256 vaultRealAssetsBefore = IERC20(asset).balanceOf(parentVault);
+        uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
+        for (uint256 i = 0; i < adaptersLength; i++) {
+            vaultRealAssetsBefore += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
+        }
+        uint256 vaultBuffer = vaultRealAssetsBefore.zeroFloorSub(IVaultV2(parentVault).totalAssets());
```
albert-llimos: `vaultBuffer` calculation calls `totalAssets()` that runs through all the adapter's `realAssets()` for a second time. `totalAssets()`  returns `min(realAssets, maxTotalAssets)`, with realAssets being the same quantity as vaultRealAssetsBefore already calculated. Then `vaultBuffer = realAssets − min(realAssets, maxTotalAssets)`.
Have you considered if it might be cheaper gas-wise to actually replicate the `maxTotalAssets` calculation instead of looping through all the adapters for the second time? Calculating it requires a single storage read from the Vault (_totalAssets, lastUpdate, maxRate are all in the same slot). It seems to me that it should be cheaper.
It's true that if the Vault has accrued interests in this transaction it might not be obviously cheaper because of the `firstTotalAssets` early return, but that's not always the case.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+        require(seller == address(this), NotSelf());
+
+        accrueInterest();
+        // forge-lint: disable-next-item(reentrancy-no-eth) updatePosition does not call back.
+        IMidnight(midnight).updatePosition(market, address(this));
+
+        uint256 vaultRealAssetsBefore = IERC20(asset).balanceOf(parentVault);
+        uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
+        for (uint256 i = 0; i < adaptersLength; i++) {
+            vaultRealAssetsBefore += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
```
albert-llimos: This adapter will be part of the list of adapters and it is currently doing an external call to itself when calling `realAssets()`. Since we know that it has accrued interests before, we could short-circuit that:

```
address adapter = IVaultV2(parentVault).adapters(i);
vaultRealAssetsBefore += adapter != address(this) ? IAdapter(adapter).realAssets() : totalAssets ;
```

I'd guess the gas savings are not much (~700 maybe?) so maybe not worth implementing for readability.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+        IVaultV2(parentVault)
+            .allocate(
+                address(this),
+                abi.encode(ids(market), boughtNetCredit.toInt256() - netCreditLoss.toInt256()),
+                paidAssets
+            );
+
+        if (timeToMaturity > 0) {
+            uint256 interest = boughtNetCredit - paidAssets;
+            uint120 growthIncrease = (interest / timeToMaturity).toUint120();
```
peyha, line 449: this could round to 0
quick magnitude check:
a 1M loan at 4% APR over a month correspond to 3k interest (`3*10^9` USDC) over ~2M6 seconds so the growth would be 1500 interest/sec.
a 20k loan at 0.1% over a day correspond to  ~0.05 interest (`5*10^4` USDC) over 86400 so the growth would round to zero
the interest being a `size*rate*ttm/num_sec_per_an`, the growth will round to zero when size*rate is small, could be worth adding a comment

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,575 @@
...
+            totalAssets += (paidAssets + interest % timeToMaturity).toUint128();
```
peyha, line 450: as a taker, when I see an open lend offer I can take such that `interest=timeToMaturity-1` several times to increase `totalAssets` directly instead of `currentGrowth`. As a taker I can deposit just before executing this 'attack' before withdrawing to seize the interest. `interest % timeToMaturity` could probably be accounted at maturity (or considered negligible)

src/adapters/MidnightAdapter.sol
adhusson: Yes, see comment:

https://github.com/morpho-org/vault-v2/blob/a43ce0b493f101270c4034e69fcc32e2d3d39c1c/src/adapters/MidnightAdapter.sol#L19

From old PR description, the issue is real with WBTC:

> Things to decide
> Scale growth by WAD? Otherwise for instance with 1 year ttm at 5%/yr, any WBTC take below ~6.3BTC is added to assets instead of being spread to growth. In general growth = interest / ttm so any take where interest < ttm the remainder goes to current assets. In the WBTC example 6.3e8 * 5/100 is ~ 1 year in seconds. But for others it's not as bad (for usdc 1 year 5%, everything goes to assets below $630).

Was closed by discussion with Mathis: only issue is these low-decimal high-value tokens, not worth moving to WAD. 

But could reopen? wdyt

src/adapters/ratifiers/MidnightAdapterEcrecoverRatifier.sol
adhusson: Having a per ratifier buffer allowance works (plus a 'take' buffer allowance) (edit: nesting problem fixed but moved to a separate PR: https://github.com/morpho-org/vault-v2/pull/996)

About changing the ratifier: there is a `take` function. The allocator can already make the adapter buy/sell at any price. So changing this ratifier does not prevent the problem.

src/adapters/MidnightAdapter.sol
peyha: I think there is indeed an issue for WBTC because I would expect rates to be much lower than that (check the markets with cbBTC on morpho blue). So large amounts of WBTC would have to be taken. not sure we want to discard these tokens so proob worth reopening imo

adhusson: https://github.com/morpho-org/vault-v2/pull/995 (@MathisGD)

src/adapters/MidnightAdapter.sol
adhusson: For high decimal tokens and/or with wad scaling this doesn't matter right?

Accounting at maturity is a bit ugly you need a separate "remainder" field.

src/adapters/MidnightAdapter.sol
adhusson: I compared the gas and you gain 10k when firstTotalAssets is not set but lose 8k when it is. You could check for it, but overall it doesn't seem worth the added code.

src/adapters/MidnightAdapter.sol
adhusson: Agree on the saving (~1k) but also agree on maybe not worth implementing.

src/adapters/MidnightAdapter.sol
albert-llimos: >   lose 8k when firstTotalAssets is set

That is higher than I would have thought but I see that we can't really get those three storage variables in one call, even if they are in the same slot.

I think there is no pathway that ends up on `onSell` and touches the Vault setting the firstTotalAssets before, no? I'd need to be a bundler transaction 🤔 

Anyway, I see how this might not be an obvious improvement, resolving.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,580 @@
...
+        SafeERC20Lib.safeTransfer(token, skimRecipient, balance);
+        emit Skim(token, balance);
+    }
+
+    /* VAULT ALLOCATORS FUNCTIONS */
+
+    function withdrawToVault(Market memory market, uint256 withdrawnAssets) external {
+        bytes32 marketId = IdLib.toId(market);
+
+        IVaultV2(parentVault).accrueInterest();
```
adhusson: Accrual is necessary here and in `onBuy`, `onSell`, and when force deallocating before a credit update. If one of those operations happens inside the callback of a buyer that buys from the adapter, the credit of the adapter has been reduced but no assets have been received yet. So if a first accrual happens, the vault would get a wrong low share price. 

The alternative to accruing the vault is to prevent nested operations by checking the liquidation lock here, in `isRatified`, in `take` and in force deallocate.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,580 @@
...
+        uint256 balance = IERC20(token).balanceOf(address(this));
+        SafeERC20Lib.safeTransfer(token, skimRecipient, balance);
+        emit Skim(token, balance);
+    }
+
+    /* VAULT ALLOCATORS FUNCTIONS */
+
+    function withdrawToVault(Market memory market, uint256 withdrawnAssets) external {
+        bytes32 marketId = IdLib.toId(market);
+        require(!IMidnight(midnight).liquidationLocked(marketId, address(this)), PositionLocked());
```
adhusson: One thing i only notice now is that in a trade where the adapter sells, the buyer callback can use the adapter again in a way that updates its credit in this market, then accrue the vault for the first time. Then the vault records total assets that are too low since it has not received the assets of the outer sell. 

Note that the problem is only at the vault level, the adapter accounting works with nested operations (it just changes roundings since all credit decreases are recorded at the inner operations).

The solution here is to prevent nested takes completely by checking the liquidation lock. 

Alternative: accrue the vault at the beginning of `onSell`, `onBuy`, `withdrawToVault` and force deallocate. Then firstTotalAssets is not zero.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,580 @@
...
+        for (uint256 i = 0; i < market.collateralParams.length; i++) {
+            address collateralToken = market.collateralParams[i].token;
+            idsArray[j++] = keccak256(abi.encode("collateralToken", collateralToken));
+            idsArray[j++] = keccak256(
+                abi.encode(
+                    "collateralParams",
+                    collateralToken,
+                    market.collateralParams[i].oracle,
+                    market.collateralParams[i].lltv,
+                    market.collateralParams[i].liquidationCursor
```
MathisGD: @claude apply
```suggestion
                    collateralParams[i]
```
we can also remove the collateralToken variable above

src/adapters/MidnightAdapter.sol
peyha: wad scaling seems to solve it yes even wiht very long duration and low decimals i don't see how the "attack" would be feasible because `timeToMaturity-1` would be impossible to reach

src/adapters/MidnightAdapter.sol
peyha: how is this a problem ? 
from what I understand, the vault will have less asset than it should so transactions might just revert because of too low erc20 balance. I'm not sure there's a risk of losing funds

src/adapters/MidnightAdapter.sol
peyha: it seems cleaner to me to have a per-market `skipBufferCheck`

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,580 @@
...
+        if (maturityData.netCredit == 0) maturityData.durationCount = uint8(durationCount(market.maturity));
+        uint256 timeToMaturity = market.maturity.zeroFloorSub(block.timestamp);
+        // current net credit cannot be > accounted net credit + bought net credit
+        uint256 netCreditLoss = marketData.netCredit + boughtNetCredit - currentNetCredit(marketId);
+        decreaseNetCredit(marketId, market.maturity, netCreditLoss);
+        uint256 idleAssets = IERC20(asset).balanceOf(parentVault);
+        if (callbackData.length > 0 && paidAssets > idleAssets) {
+            (address fundingAdapter, bytes memory fundingData) = abi.decode(callbackData, (address, bytes));
+            // forge-lint: disable-next-item(reentrancy-no-eth) the adapter is trusted.
+            IVaultV2(parentVault).deallocate(fundingAdapter, fundingData, paidAssets - idleAssets);
```
peyha: isn't this part missing a few cases ? 
for instance it's impossible to withdrawtovault before buying 
not sure how it could be generalized (maybe through a self call)

adhusson: good point, will add it, do you have other cases in mind?

peyha: mmh to fetch liquidity at this moment, the adapter could
- deallocate on another adapter 
- deallocate on this adapter by either 1. withdrawToVault 2. sell (might not be relevant)
I don't think there would be any other case

src/adapters/MidnightAdapter.sol
adhusson: Example:

```
midnight.take(adapter sells 4 credit for 4 assets)
- Midnight credit: 10 -> 6
- attacker.onBuy
  - forceDeallocate(0)
    - adapter cache: 10 -> 6
    - vault accrues: totalAssets = 6
  - deposit 6 assets -> receive 100 shares
 ```

The deposit happened with a small share price that should not be visible.

File: src/adapters/MidnightAdapter.sol (hunk 43 above)
Comment (adhusson) [reply src/adapters/MidnightAdapter.sol]: self-funding done

File: src/adapters/MidnightAdapter.sol (hunk 38 above)
Comment (adhusson) [reply src/adapters/MidnightAdapter.sol:427]: closing we can keep going in https://github.com/morpho-org/vault-v2/pull/786#discussion_r3980918609

File: test/MidnightAdapterTest.sol [hunk 44]
```diff
@@ -0,0 +1,2998 @@
...
+        Offer memory offer = makeBuyOffer(30 days, 1e18, MAX_TICK);
+        midnight.supplyCollateral(offer.market, 0, offer.maxUnits, taker);
+        bytes memory data = sign([offer], signerAllocator);
+        assertEq(adapter.minRate(), 0, "default minRate");
+        setMinRate(1);
+
+        vm.prank(taker);
+        vm.expectRevert(IMidnightAdapter.RateTooLow.selector);
+        midnight.take(offer, data, offer.maxUnits, taker, taker, address(0), "");
+        assertEq(adapter.totalAssets(), 0, "failed buy leaves no assets");
```
peyha: the test cannot compile because adapter (which is `IMidnightAdapter`) does not have `.totalAssets()`

peyha:
```suggestion
        assertEq(adapter.realAssets(), 0, "failed buy leaves no assets");
```

src/adapters/MidnightAdapter.sol
adhusson: firstTotalAssets prevents looping this in one tx. For multiple tx maxrate protects as well

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,525 @@
...
+        SafeERC20Lib.safeTransfer(token, skimRecipient, balance);
+        emit Skim(token, balance);
+    }
+
+    /* VAULT ALLOCATORS FUNCTIONS */
+
+    function withdrawToVault(Market memory market, uint256 withdrawnAssets) external {
+        bytes32 marketId = IdLib.toId(market);
+
+        require(!IMidnight(midnight).liquidationLocked(marketId, address(this)), SellInProgress());
```
peyha, line 258: why not put this check in `deallocate` ? you could also remove the one in `isRatified` and in the `forceDeallocate` part this way

adhusson: `deallocate` is called by `onSell`, where the lock is active, so it would revert (and we don't want to revert). Also, inside `onBuy` the taker could take again such that deallocate is not called, so the check in `isRatified` would still be necessary.

peyha: oh yes good point for the revert

peyha: I was wondering if it was fine to let the adapter buy within the taker `onBuy` because it would necessary call `allocate` and `accrueInterest` on the vault (which would either revert because of the `realAssets` check or use a stall interest state which should be fine)

adhusson: yes buying in the same market will revert, but buying in another market can work if the vault has already accrued before the take in this transaction

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,515 @@
...
+        require(boughtNetCredit >= paidAssets, BuyAtLoss());
+
+        // Cache corrected net credit before call to allocate
+        uint128 newNetCredit = currentNetCredit(marketId, market);
+        (overridenMarketId, overridenMarketNetCredit) = (marketId, newNetCredit - boughtNetCredit);
+        IVaultV2(parentVault).accrueInterest();
+        (overridenMarketId, overridenMarketNetCredit) = (bytes32(0), 0);
+
+        if (block.timestamp < market.maturity && boughtNetCredit > 0) {
+            uint256 boughtGrowth =
```
adhusson: Moved to storing market assets [growth](https://github.com/morpho-org/vault-v2/pull/786/changes/b1707672c7afe564c5877932f7a838959c9a3fb0) instead of storing current assets + lastUpdate. Markets use 1 less slot, and there is nothing to update on sells/losses except the credit.

Growth is in 1/WAD per second of the net credit. There is a growth rounding down that moves credit to assets too early, but it has a much smaller impact than with growth in assets per second:

- If the adapter buys `c` credit for `a` assets the growth will be `floor((c-a) * WAD  / (c*ttm))`, so the immediate asset increase is at most c*ttm/WAD.
- With repeated buys the rounding repeats. So if there is a borrow on 5 years with total debt $150M, the rounding will overaccrue at most $0.0236. After 1e4 buys that repeat the rounding (they can be 1 wei buys) at most $236 has been accrued too early.

MathisGD: note that it was possible to make it fit in one slot:
image

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,515 @@
...
+    mapping(bytes4 selector => uint256) public timelock;
+    mapping(bytes4 selector => bool) public abdicated;
+    mapping(bytes data => uint256) public executableAt;
+
+    /* MANAGEMENT */
+
+    address public skimRecipient;
+    bool public skipBufferCheck;
+    /// @dev Minimum net simple interest rate per second, WAD-scaled, enforced on maker and taker buys before maturity.
+    uint256 public minRate;
```
peyha, line 56: not sure I understand this concept, isn't it equivalent to having a max tick/max price when buying ?

adhusson: [context](https://morpholabs.slack.com/archives/C08A40KKN9W/p1788960507663579)

it prevents allocators from signing huge buy offers at 0%

src/adapters/MidnightAdapter.sol
adhusson: With uint80 remaining interest and 18 decimals remaining interest is at most ~1e6 tokens. `maturity` could be a uint32 with a shift from 2026 always applied, so remainingInterest becomes a `uint88` and now it's ~256e6. Still a bit aggressive.

src/adapters/interfaces/IMidnightAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+
+import {IAdapter} from "../../interfaces/IAdapter.sol";
+import {Market, Offer} from "lib/midnight/src/interfaces/IMidnight.sol";
+import {IBuyCallback, ISellCallback} from "lib/midnight/src/interfaces/ICallbacks.sol";
+import {IRatifier} from "lib/midnight/src/interfaces/IRatifier.sol";
+
+struct MarketData {
+    uint128 netCredit;
+    /// @dev Each unit of growth represents 1/WAD of a raw asset unit accrued per second, until maturity.
+    uint192 growth;
```
MathisGD: right?
```suggestion
    uint64 growth;
```

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,514 @@
...
+
+        MaturityData storage maturityData = _maturities[market.maturity];
+        // forge-lint: disable-next-item(unsafe-typecast) durationCount <= MAX_DURATIONS.
+        if (maturityData.netCredit == 0) maturityData.durationCount = uint8(durationCount(market.maturity));
+        int256 change = updateNetCredit(marketId, market, newNetCredit);
+        uint256 idleAssets = IERC20(asset).balanceOf(parentVault);
+        if (callbackData.length > 0 && paidAssets > idleAssets) {
+            (address fundingAdapter, bytes memory fundingData) = abi.decode(callbackData, (address, bytes));
+            if (fundingAdapter == address(this)) {
+                this.withdrawToVault(abi.decode(fundingData, (Market)), paidAssets - idleAssets);
```
MathisGD: doesn't need to be a self call right? could be an internal call

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,514 @@
...
+        }
+
+        MaturityData storage maturityData = _maturities[market.maturity];
+        // forge-lint: disable-next-item(unsafe-typecast) durationCount <= MAX_DURATIONS.
+        if (maturityData.netCredit == 0) maturityData.durationCount = uint8(durationCount(market.maturity));
+        int256 change = updateNetCredit(marketId, market, newNetCredit);
+        uint256 idleAssets = IERC20(asset).balanceOf(parentVault);
+        if (callbackData.length > 0 && paidAssets > idleAssets) {
+            (address fundingAdapter, bytes memory fundingData) = abi.decode(callbackData, (address, bytes));
+            if (fundingAdapter == address(this)) {
```
MathisGD, line 413: btw I don't think that this feature is important, or maybe I'm missing something

adhusson: [context](https://github.com/morpho-org/vault-v2/pull/786#discussion_r4006700137), in rolling a taker could manually withdrawToVault then use the buy offer yes.

peyha: I also think it would be helpful for rolling

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    function setMinRate(uint256 newMinRate) external {
+        timelocked();
+        minRate = newMinRate;
+        emit SetMinRate(newMinRate);
+    }
+
+    /// @dev Help prevent operational errors when selling.
+    function setMinSellPrice(bytes32 marketId, uint256 newMinSellPrice) external {
```
peyha: not timelocked?

adhusson: the idea is that its to help with operational errors but is not a defense against attacks

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,514 @@
...
+
+    function ids(Market memory market) public view returns (bytes32[] memory) {
+        uint256 durationsCount = _maturities[market.maturity].durationCount;
+
+        bytes32[] memory idsArray = new bytes32[](2 + market.collateralParams.length * 2 + durationsCount);
+
+        uint256 j;
+        idsArray[j++] = adapterId;
+        idsArray[j++] =
+            keccak256(abi.encode("marketConfig", market.enterGate, market.liquidatorGate, market.rcfThreshold));
```
peyha, line 509: shouldn't we add a dedicated id for `enterGate` ? I could imagine a vault wanting to have a cap for markets with users whitelisted by some country/company

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,518 @@
...
+        Market memory dummyMarket;
+        for (uint256 i = 0; i < length; i++) {
+            bytes32 marketId = marketIds[i];
+            MarketData memory marketData = _markets[marketId];
+            uint256 newNetCredit;
+            if (marketId == overridenMarketId) {
+                newNetCredit = overridenMarketNetCredit;
+            } else {
+                require(!IMidnight(midnight).liquidationLocked(marketId, address(this)), SellInProgress());
+                newNetCredit = currentNetCredit(marketId, dummyMarket);
```
MathisGD: the reason why sending a fully empty dummyMarket here is ok isn't that straightforward. we could send a market with the right maturity at least (only the maturity is read, too bad btw that we didn't make the interface simpler)

adhusson: it saves only ~300 gas per market, we can add it back

src/adapters/MidnightAdapter.sol
adhusson: not sure either way, @MathisGD ?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,520 @@
...
+        (overridenMarketId, overridenMarketNetCredit) = (0, 0);
+
+        if (block.timestamp < market.maturity && boughtNetCredit > 0) {
+            uint256 boughtGrowth = (boughtNetCredit - paidAssets).mulDivDown(WAD, market.maturity - block.timestamp);
+            require(boughtGrowth >= minRate * paidAssets, RateTooLow());
+
+            MarketData storage marketData = _markets[marketId];
+            uint256 remainingGrowth = (newNetCredit - boughtNetCredit) * marketData.growth;
+            // forge-lint: disable-next-item(unsafe-typecast) growth <= WAD < 2**64.
+            marketData.growth = uint64((remainingGrowth + boughtGrowth) / newNetCredit);
```
peyha: not sure I totally understand your computation
I would expect the new total interest `newGrowth * ttm` to be equal to `previousInterest + addedInterest` where `previousInterest=growth*ttm` and `addedInterest=(boughtNetCredit-paidAssets)` ultimately giving `newGrowth=(growth*ttm+boughtNetCredit-paidAssets)/ttm` 

could also be equivalent to `newGrowth=growth+(boughtNetCreidt-paidAssets)/ttm`

peyha: ah I just understood that the definition of growth changed and is now scaled by credit
I have the same formula

adhusson: `marketData.growth` is now in wad units of net credit! so 1 of growth means every second the amortized value increases by 1e-18th of the net credit. I'll improve the comment in the MarketData struct and rename the `remainingGrowth` and `boughtGrowth` variables.

src/adapters/MidnightAdapter.sol
MathisGD: what saves 300 gas exactly

MathisGD: ok got the answer in the other PR. applying there

src/adapters/MidnightAdapter.sol
adhusson: i think we are not doing an absolute amount buffer anyway, ok to resolve?

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,522 @@
...
+
+    mapping(bytes4 selector => uint256) public timelock;
+    mapping(bytes4 selector => bool) public abdicated;
+    mapping(bytes data => uint256) public executableAt;
+
+    /* MANAGEMENT */
+
+    address public skimRecipient;
+    /// @dev Minimum net simple interest rate per second, WAD-scaled, enforced on maker and taker buys before maturity.
+    uint256 public minRate;
```
albert-llimos: I think renaming this to `minBuyRate ` would make it more clear what it is.

src/adapters/MidnightAdapter.sol
```diff
@@ -0,0 +1,533 @@
...
+
+        packedDurations = DurationsLib.pack(_durations);
+        durationsLength = _durations.length;
+    }
+
+    /* GETTERS */
+
+    function netCredit(bytes32 marketId) external view returns (uint128) {
+        return _markets[marketId].netCredit;
+    }
```
MathisGD: there should be a getter per element

src/adapters/ratifiers/MidnightAdapterSetterRatifier.sol
adhusson: will add rate ratifier after

src/adapters/MidnightAdapter.sol
peyha: me neither, the code must have changed (unless I was saying nonsense)

Issue #793, opened 2025-11-14
Struct interface of market adapter


QGarchery:
See [this thread](https://morpholabs.slack.com/archives/C08A40KKN9W/p1760601825553819)

PR #789, opened 2025-11-13
MarketV1AdapterV2

MathisGD:
- add the force remove feature (timelocked)
- add shares accounting (for full donation resistance)
- add events in allocate/deallocate to allow to easily rebuild the market list (see [this](https://morpholabs.slack.com/archives/C08A40KKN9W/p1760455137830219))
- gas improvement: compute `expectedTotalAssets` without passing MarketParams as input (see #816)

PR #794, opened 2025-11-14
Fix interface of market adapter

QGarchery:
If we do #790 & #789, this can be merged

Fixes:
- #793


MathisGD:
not useful anymore since we go for #790

PR #796, opened 2025-11-17
suggestions on top of #790

MathisGD:


src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -130,16 +132,9 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
             MarketParams({loanToken: loanToken, collateralToken: collateralToken, oracle: oracle, irm: irm, lltv: lltv});
     }
 
-    function realAssets() external view returns (uint256) {
-        return allocation() != 0 ? expectedSupplyAssets() : 0;
```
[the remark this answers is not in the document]
adhusson: the comparison [here](https://github.com/morpho-org/vault-v2/pull/790#discussion_r2527858676) takes it into account, and the third column ('pack isAllocated bool') tries to reduce that cost.

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -130,16 +132,9 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
-    }
-
-    /* INTERNAL FUNCTIONS */
-
-    function expectedSupplyAssets() internal view returns (uint256) {
+    function realAssets() public view returns (uint256) {
```
adhusson: If we completely remove the `allocation == 0` check, then a market with leftover shares but that is completely deallocated could later be called. So an emergency deallocation would not completely prevent the vault from calling the market.

The current MarketV1 adapter [prevents this](https://github.com/morpho-org/vault-v2/blob/main/src/adapters/MorphoMarketV1Adapter.sol#L113), and the current VaultV1 adapter [prevents this as well](https://github.com/morpho-org/vault-v2/blob/main/src/adapters/MorphoVaultV1Adapter.sol#L111).

MathisGD: why don't we want to avoid calling the market? if it reverts, then you would do a removeAdapter, right?

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -21,43 +20,43 @@ import {SafeERC20Lib} from "../libraries/SafeERC20Lib.sol";
...
-    using MarketParamsLib for MarketParams;
-    using SharesMathLib for uint256;
+    using SharesMathLib for uint128;
 
     /* IMMUTABLES */
 
     address public immutable factory;
     address public immutable parentVault;
     address public immutable morpho;
+    address private immutable loanToken;
```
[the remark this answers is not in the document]
adhusson: you can already read it in `marketParams()`

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -83,14 +82,21 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
 
-        // Safe casts because Market V1 bounds the total supply of the underlying token, and allocation is less than the
-        // max total assets of the vault.
-        return (ids(), int256(expectedSupplyAssets()) - int256(allocation()));
+        uint256 newAllocation = realAssets();
+        // Safe casts because Market V1 bounds totalSupplyAssets to uint128.max.
+        int256 change = int256(newAllocation) - int256(uint256(allocation));
+        allocation = uint128(newAllocation);
+
+        emit Allocate(marketParams(), newAllocation, shares);
```
peyha: it's weird from an indexing POV to log `newAllocation` which is a storage value with `shares` which is the difference between the previous and current storage value, we could harmonize it

PR #809, opened 2025-11-19
[force remove] share price check for #790

peyha:
As discussed this morning, implements the same thing as #408


adhusson (approved):

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -87,6 +87,10 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
         uint256 mintedShares;
         if (assets > 0) {
             (, mintedShares) = IMorpho(morpho).supply(marketParams(), assets, 0, address(this), hex"");
+
+            // Rounding error protection
```
MathisGD:
```suggestion
```

PR #810, opened 2025-11-19
[force remove] force remove market

adhusson:


test/MorphoMarketV1AdapterTest.sol
```diff
@@ -327,4 +331,116 @@ contract MorphoMarketV1AdapterTest is Test {
...
+        adapter.submitForceRemove();
+
+        uint256 snap = vm.snapshotState();
+
+        vm.prank(curator);
+        vm.expectEmit();
+        emit IMorphoMarketV1Adapter.RevokeForceRemove();
+        adapter.revokeForceRemove();
+
+        assertEq(adapter.forceRemoveExecutableAt(), 0);
```
QGarchery: What's the point of the snapshot, why do we test it twice here ?

QGarchery:
some slight changes in tests in #815

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -78,6 +79,31 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
...
+        forceRemoveExecutableAt = 0;
+        emit RevokeForceRemove();
+    }
+
+    function forceRemove() external {
+        require(forceRemoveExecutableAt != 0, NotTimelocked());
+        require(block.timestamp >= forceRemoveExecutableAt, TimelockNotExpired());
+        supplyShares = 0;
+        forceRemoveExecutableAt = 0;
+        emit ForceRemove();
```
peyha: could log the number of burnt shares for the accounting

PR #814, opened 2025-11-20
[force remove] market params store

adhusson:
[updated comparison sheet](https://docs.google.com/spreadsheets/d/19e2etXgO_KYp6MvajuMF4EVlfWeuy-viPYN1yiXVI7A/edit?usp=sharing)


adhusson:
removing in favor of https://github.com/morpho-org/vault-v2/pull/816

note for later: 
* could go further and have a single store contract, redeployed at every market list change (up to redeploy cost, to check)
* could avoid having to store marketparams in liquidity adapter data
* even further could use create3 to map uint16 indices to store contracts, so markets in the list are packed by 256/16=16 markets per slot.

PR #816, opened 2025-11-20
[force remove] gas savings with adaptive curve irm only

adhusson:


.gitmodules
```diff
@@ -12,4 +12,7 @@
 	url = https://github.com/morpho-org/metamorpho
 [submodule "lib/metamorpho-v1.1"]
 	path = lib/metamorpho-v1.1
-	url = https://github.com/morpho-org/metamorpho-v1.1
+	url = git@github.com:morpho-org/metamorpho-v1.1.git
```
MathisGD: why?

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -35,26 +37,29 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1AdapterStaticTyping {
     address public immutable asset;
     address public immutable morpho;
     bytes32 public immutable adapterId;
+    address public immutable irm;
```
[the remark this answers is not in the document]
adhusson: depends on the name of the contract, if it's "AdaptiveCurveMarketAdapter" it can stay `irm`

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -157,25 +163,25 @@ contract MorphoMarketV1Adapter is IMorphoMarketV1AdapterStaticTyping {
         return (ids(marketParams), int256(_newAllocation) - int256(oldAllocation));
     }
 
-    function updateList(MarketParams memory marketParams, uint256 oldAllocation, uint256 _newAllocation) internal {
+    function updateMarketIds(Id marketId, uint256 oldAllocation, uint256 _newAllocation) internal {
```
MathisGD:
```suggestion
    function updateList(Id marketId, uint256 oldAllocation, uint256 _newAllocation) internal {
```

src/adapters/interfaces/IMorphoMarketV1Adapter.sol
```diff
@@ -3,30 +3,47 @@
 pragma solidity >=0.5.0;
 
 import {IAdapter} from "../../interfaces/IAdapter.sol";
-import {Id, MarketParams} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+import {Id, MarketParams} from "../../../lib/morpho-blue-irm/lib/morpho-blue/src/interfaces/IMorpho.sol";
```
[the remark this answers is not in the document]
MathisGD: >until we update morpho-blue.

wdym

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "lib/metamorpho-v1.1"]
 	path = lib/metamorpho-v1.1
 	url = https://github.com/morpho-org/metamorpho-v1.1
+[submodule "lib/morpho-blue-irm"]
+	path = lib/morpho-blue-irm
+	url = https://github.com/morpho-org/morpho-blue-irm
```
QGarchery: I think we should remove the morpho-blue module now

src/adapters/interfaces/IMorphoMarketV1Adapter.sol
```diff
@@ -38,27 +37,16 @@ interface IMorphoMarketV1AdapterBase is IAdapter {
     function parentVault() external view returns (address);
     function asset() external view returns (address);
     function morpho() external view returns (address);
-    function positions(Id id) external view returns (uint128 supplyShares, uint128 allocation);
+    function marketIds(uint256 index) external view returns (Id);
+    function positions(Id marketId) external view returns (uint128, uint128);
```
QGarchery: If we are not going to return a `MarketPosition` struct, then we might as well move the definition of that struct to the implementation

src/adapters/MorphoMarketV1Adapter.sol
```diff
@@ -25,6 +28,7 @@ import {SafeERC20Lib} from "../libraries/SafeERC20Lib.sol";
 /// has zero shares on the market.
 contract MorphoMarketV1Adapter is IMorphoMarketV1Adapter {
```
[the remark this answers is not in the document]
adhusson: will do after merge

src/adapters/interfaces/IMorphoMarketV1Adapter.sol
adhusson: I see your point, but it's also about "where to go to get definitions". It's nice to know in advance you open `I<something>` to get the structs.

adhusson: 2 getters vs returning `(uint128, uint128)` doesn't go one way or the other wrt where `MarketPosition` should be defined.

It's lower cost to do 1 call & throw away a value than to do 2 calls and not too ugly in code so I would keep the tuple.

MathisGD: let's return the struct actually, it's not very costly

MathisGD: #822

PR #822, opened 2025-11-22
refactor: market adapter interface

MathisGD:


src/adapters/interfaces/IMorphoMarketV1Adapter.sol
```diff
@@ -50,3 +49,11 @@ interface IMorphoMarketV1Adapter is IAdapter {
...
     function skim(address token) external;
 }
+
+interface IMorphoMarketV1AdapterStaticTyping is IMorphoMarketV1AdapterBase {
+    function positions(Id marketId) external view returns (uint128 supplyShares, uint128 allocation);
+}
+
+interface IMorphoMarketV1Adapter is IMorphoMarketV1AdapterBase {
+    function positions(Id marketId) external view returns (MarketPosition memory);
+}
```
adhusson: Isn't this way too complex for just returning 2 values?

MathisGD: kind of convincing

PR #825, opened 2025-11-24
Fix verif and modules

QGarchery:


foundry.lock
```diff
@@ -9,13 +9,10 @@
     "rev": "2d160ba9bb945ca3bf12efb182427445dce59c27"
   },
   "lib/morpho-blue": {
-    "tag": {
-      "name": "v1.0.0",
-      "rev": "55d2d99304fb3fb930c688462ae2ccabb1d533ad"
-    }
+    "rev": "cf3f0ce68db99421bcd808d505cfe49d61f4eaa0"
```
QGarchery: it was downgraded in the target PR

PR #839, opened 2025-12-01
refactor timelocks 1

MathisGD:
- refactor the marketV1AdapterV2 to have the same timelock system as the vault itself
- also changes the timelock of burnShares to its own one
- also make setSkimRecipient timelocked

PR #842, opened 2025-12-02
refactor timelocks 2

MathisGD:
alternative to #839

major drawbacks: 
- the adapter needs to be sentinel
- there can be a selector clash


MathisGD:
replaced by #839

PR #849, opened 2025-12-03
Re-order errors alphabetically

peyha:
Was a bit annoying to check the consistency of the new adapter and the vault errors because of this one. From what I tested, it does not change the bytecode of vaults


peyha:
Tried checking that the new timelock errors from `IMorphoMarketV1AdapterV2.sol` were the same as in `ErrorsLib` and noticed this inconsistency

MathisGD (approved):

PR #863, opened 2025-12-10
Show counter example in the vault adapter factory

QGarchery:


QGarchery:
Closing as this was only useful as an example for #857

PR #872, opened 2025-12-23
fixed earliest-executable-time formula in the doc.  Now consistent with the certora spec

bhargavbh:
trivial fix


MathisGD (approved):
nice catch!

PR #873, opened 2025-12-30
[Certora] prove revert equivalence for transfer and transferFrom

bhargavbh:
Added rules for `transfer` and `transferFrom` Revert Conditions. 

Pending due to Certora issues:
- `submit` 
- `setName`
- `setSymbol`
- `setLiquidityAdapterAndData`

WIP Verification efforts in this PR- https://github.com/morpho-org/vault-v2/pull/868
- `allocate`
- `deallocate`

### DIFF
--- a/certora/specs/Reverts.spec
+++ b/certora/specs/Reverts.spec
@@ -6,8 +6,6 @@ import "../helpers/UtilityVault.spec";
 using RevertCondition as RevertCondition;
 using Utils as Utils;
 
-//definition MAX_UINT256() returns uint256 = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
-
 // This specification checks either the revert condition or the input validation under which a function reverts.
 // Interest accrual is assumed to not revert.
 


certora/specs/Reverts.spec
```diff
@@ -176,21 +178,29 @@ rule setMaxRateRevertCondition(env e, uint256 newMaxRate) {
...
 }
 
-rule transferInputValidation(env e, address to, uint256 shares) {
+rule transferRevertCondition(env e, address to, uint256 shares) {
     bool toIsZeroAddress = to == 0;
     bool callerCanSendShares = canSendShares(e.msg.sender);
     bool toCanReceiveShares = canReceiveShares(to);
 
+    require shares + balanceOf(to) <= MAX_UINT256(); // To avoid overflow in the balanceOf check.
+    require shares <= balanceOf(e.msg.sender); // To avoid underflow in the balanceOf check.
```
MathisGD: let's write these as elements of the or in the assert line (to have all the requirements at he same place)

certora/specs/Reverts.spec
```diff
@@ -176,21 +178,29 @@ rule setMaxRateRevertCondition(env e, uint256 newMaxRate) {
...
+rule transferFromRevertCondition(env e, address from, address to, uint256 shares) {
     bool fromIsZeroAddress = from == 0;
     bool toIsZeroAddress = to == 0;
     bool fromCanSendShares = canSendShares(from);
     bool toCanReceiveShares = canReceiveShares(to);
+    bool sharesLessEqualAllowance = e.msg.sender != from => (shares <= allowance(from, e.msg.sender)); // To avoid underflow in the allowance check.
+
+    require shares + balanceOf(to) <= MAX_UINT256(); // To avoid overflow in the balanceOf check.
+    require shares <= balanceOf(e.msg.sender); // To avoid underflow in the balanceOf check.
+    require shares <= balanceOf(from); // To avoid underflow in the balanceOf check.
```
MathisGD: same remark

certora/specs/Reverts.spec
MathisGD: more than having them all at the same place actually, here we are not verifying the equivalence (see [this bug](https://github.com/morpho-org/vault-v2/pull/873/changes#r2653486098))

certora/specs/Reverts.spec
```diff
@@ -176,21 +178,29 @@ rule setMaxRateRevertCondition(env e, uint256 newMaxRate) {
...
+    assert toIsZeroAddress || !callerCanSendShares || !toCanReceiveShares || e.msg.value != 0 <=> lastReverted;
 }
 
-rule transferFromInputValidation(env e, address from, address to, uint256 shares) {
+rule transferFromRevertCondition(env e, address from, address to, uint256 shares) {
     bool fromIsZeroAddress = from == 0;
     bool toIsZeroAddress = to == 0;
     bool fromCanSendShares = canSendShares(from);
     bool toCanReceiveShares = canReceiveShares(to);
+    bool sharesLessEqualAllowance = e.msg.sender != from => (shares <= allowance(from, e.msg.sender)); // To avoid underflow in the allowance check.
```
MathisGD: let's try to be more high level in the comments
```suggestion
    bool sharesLessEqualAllowance = e.msg.sender != from => (shares <= allowance(from, e.msg.sender)); // msg.sender should have enough allowance.
```

certora/specs/Reverts.spec
bhargavbh: good point! 
agree that using require is not logically equivalent to declaring a boolean and adding it as a condition in the assert statement. The proofs fail with the proposed changes, i.e., !balanceWontOverflow is true but lastReverted is false. This happens when `to == e.msg.sender`.   So the condition should be modified into `to!= e.msg.sender => shares + balanceOf(to) <= MAX_UINT256()`

certora/specs/Reverts.spec
bhargavbh: addressed. similarly, the balanceWontOverflow condition here is `to != from => shares + balanceOf(to) <= MAX_UINT256()`

bhargavbh:
addressed the review comments. The boolean conditions now have descriptive naming, hence comments as such is not needed.

QGarchery (approved):

certora/specs/Reverts.spec
```diff
@@ -176,21 +178,26 @@ rule setMaxRateRevertCondition(env e, uint256 newMaxRate) {
...
-rule transferFromInputValidation(env e, address from, address to, uint256 shares) {
+rule transferFromRevertCondition(env e, address from, address to, uint256 shares) {
     bool fromIsZeroAddress = from == 0;
     bool toIsZeroAddress = to == 0;
     bool fromCanSendShares = canSendShares(from);
     bool toCanReceiveShares = canReceiveShares(to);
-
+    bool sufficientAllowance = e.msg.sender != from => (shares <= allowance(from, e.msg.sender));
+    bool balanceWontOverflow = to != from => shares + balanceOf(to) <= MAX_UINT256();
+    bool balanceWontUnderflow = shares <= balanceOf(from);
```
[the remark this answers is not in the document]
bhargavbh: shouldn't this be:
balanceWontUnderflow -> sufficientBalance
balanceWontOverflow -> boundedBalance

Issue #878, opened 2026-01-08
adapterSupplySharesIsLessThanActualSupplyShares rule fails when run on Certora-8.6.3


bhargavbh:
[...]

PR #879, opened 2026-01-12
[Certora] initialization of contracts

QGarchery:
Fixes:
- #878 

The new release is now more conservative and doesn't assume that all contracts are 0 at the start, only the currentContract.  Unfortunately, this makes the CI fail (see corresponding issue above) because, in the run:
- the currentContract is the vault v2, not the adapter
- which means that it can choose an arbitrary value for the starting internal accounting of shares. In particular it can choose a number of accounted shares that is greater than the number of shares it actually holds

2 fixes are possible:
- separate this invariant in another file, where the current contract is the adapter.
- just use the previous settings of the prover, which the provided flag. This is the approach of this PR, chosen for the simplicity. [Here](https://prover.certora.com/output/66603/21e9a49251184ea2aed02c948fde4e97) is a passing run

### DIFF
--- a/certora/confs/AllocationMorphoMarketV1AdapterV2.conf
+++ b/certora/confs/AllocationMorphoMarketV1AdapterV2.conf
@@ -24,6 +24,7 @@
   "loop_iter": "4",
   "exclude_rule": "allocationIsInt256",
   "prover_args": [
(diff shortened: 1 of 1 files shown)


bhargavbh (approved):
lgtm!

PR #900, opened 2026-03-03
Rename morpho market v1 -> morpho blue

peyha:
By mining new addresses to keep the same bytecode
Tackling #899 
Already quite convincing we should not do it because the event name cannot be changed as event signature cannot be realistically mined

Did not touch any certora file


README.md
```diff
@@ -19,7 +19,7 @@ This is notably useful when abdicated (see [timelocks](#timelocks)), to ensure t
 
 The following adapters are currently available:
 
-- [Morpho Market V1 Adapter V2](./src/adapters/MorphoMarketV1AdapterV2.sol).
+- [Morpho Blue Adapter V2](./src/adapters/MorphoBlueAdapterV2.sol).
```
MathisGD:
```suggestion
- [Blue Adapter V2](./src/adapters/BlueAdapterV2.sol).
```

PR #910, opened 2026-04-16
[midnight adapter] `forceDeallocate` by minting shares

adhusson:
In #786, if the obligation has a trading fee a user cannot do IKR without having the funds to pay the trading fee.

In this PR, IKR is free because the user gets shares in the adapter position. He is then safe from any vault attack and can sell his obligation shares later.

It also removes the question of whether to allow offer callbacks when a user force deallocates (because in the other PR, force deallocate triggers a take)


src/adapters/MidnightAdapter.sol
```diff
@@ -22,6 +20,9 @@ import {DurationsLib} from "./libraries/DurationsLib.sol";
 /// @dev The adapter must have the allocator role in its parent vault to buy, and the allocator or sentinel role to
 /// make sell offers and to withdraw to the vault.
 /// @dev If the parent vault has a sendSharesGate, the gate must allow the adapter to send shares.
+/// @dev Force deallocators get shares of the adapter's position instead of triggering a market sale. Their claims must
+/// stay redeemable even if the vault removes the adapter or its allocator role, so withdrawShares never interacts with
+/// the parent vault.
```
adhusson:
```suggestion
/// @dev Force deallocators get shares of the adapter's position instead of triggering a market sale. Their claims must
/// stay redeemable even if the vault removes the adapter, so withdrawShares never interacts with the parent vault.
```

src/adapters/MidnightAdapter.sol
```diff
@@ -120,33 +124,62 @@ contract MidnightAdapter is IMidnightAdapter {
...
+        unreportedVaultDecrease[marketId] = 0;
         IVaultV2(parentVault)
-            .deallocate(address(this), abi.encode(ids(market), -netCreditDecrease.toInt256()), withdrawnAssets);
-        emit WithdrawToVault(marketId, withdrawnAssets, netCreditDecrease);
+            .deallocate(address(this), abi.encode(ids(market), -reportedDecrease.toInt256()), withdrawnAssets);
+        emit WithdrawToVault(marketId, withdrawnAssets, reportedDecrease);
+    }
+
+    /// @dev Does not interact with the parent vault, so that claims stay redeemable even if the adapter has been
+    /// removed from the vault or has lost its allocator role.
```
adhusson:
```suggestion
    /// @dev Does not interact with the parent vault, so that claims stay redeemable even if the adapter has been
    /// removed from the vault.
```

adhusson:
removed in favor of letting the user find enough loan tokens to pay the settlement fee when there is no liquidity on midnight

PR #919, opened 2026-05-05
fix CI forge 1.7

QGarchery:


foundry.toml
```diff
@@ -12,11 +12,10 @@ wrap_comments = true
 [lint]
 exclude_lints = [
     "screaming-snake-case-immutable",
-    "asm-keccak256"
-]
-ignore = [
-    'test/**/*.sol'
+    "asm-keccak256",
+    "block-timestamp",
```
MathisGD, line 16: Ref: 
> block-timestamp — Flags use of block.timestamp as an operand of a comparison, where its value can be slightly manipulated by the block proposer

MathisGD (approved):

adhusson (approved):

PR #921, opened 2026-05-20
[midnight adapter] unordered array

adhusson:
Simpler to think about than linked list code.

For gas there is an extra cost of having to open <= 10 additional slots over the lifetime of the contract (the max slots used by the array). 

Then we get those gas numbers:

```
approx. average number of cold reads

operation                         linked list             unsorted
----------------------------------------------------------------------
accrual, == 0 maturities reached  ~1                     ~1
accrual, >= 1 maturity reached    ~reached               ~10 + reached
insert new maturity               ~future maturities/2   ~1
delete maturity to 0              ~1                     ~1
```


src/adapters/interfaces/IMidnightAdapter.sol
```diff
@@ -7,14 +7,11 @@ import {Market} from "lib/midnight/src/interfaces/IMidnight.sol";
...
 
-// Chain of maturities, each can represent multiple markets.
-// nextMaturity is 0 if no next maturity.
 struct MaturityData {
     uint128 netCredit;
     uint128 growth;
-    uint48 prevMaturity;
-    uint48 nextMaturity;
     uint8 durationCount;
+    uint8 index; // index in pendingMaturities
```
adhusson, line 14: We could have `uint120 growth`, pack `growth` with `durationCount`, remove `index`, and scan the pending maturities every time we remove one (max 22k gas) to get 1 less slot used per maturity.

Done here https://github.com/morpho-org/vault-v2/pull/932

PR #923, opened 2026-05-26
whitelist send assets gate

MathisGD:
- simple send asset gate (the non critical one)
- setter and setterWithSig (one can imagine the user itself passing the signature before interacting with the vault)
- multi-call to set multiple users at once
- intermediary feature for bundles

### DIFF
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -17,8 +17,8 @@ import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
 contract WhitelistSendAssetsGate is IWhitelistSendAssetsGate {
     address public whitelister;
     mapping(address => uint256) public nonces;
-    mapping(address => bool) public isIntermediary;
     mapping(address => bool) public isWhitelisted;
+    mapping(address => bool) public isIntermediary;
 
     constructor(address _whitelister) {
         whitelister = _whitelister;
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -71,7 +71,7 @@ contract WhitelistSendAssetsGate is IWhitelistSendAssetsGate {
         bytes32 r,
         bytes32 s
     ) external {
-        require(deadline >= block.timestamp, PermitDeadlineExpired());
+        require(deadline >= block.timestamp, DeadlineExpired());
         bytes32 hashStruct =
             keccak256(abi.encode(SET_IS_WHITELISTED_TYPEHASH, account, newIsWhitelisted, nonces[account]++, deadline));
         bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), hashStruct));
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -10,8 +10,10 @@ import {
 import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
 
 /// @dev Using this gate allows to restrict who the funds are initially owned by in a vault's deposits/mints.
-/// @dev As with any send assets gate, nothing prevents whitelisted accounts from using a non whitelisted account's funds.
-/// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead. Thus the intermediary should only deposit and mint using assets initially owned by its current initiator.
+/// @dev As with any send assets gate, nothing prevents whitelisted accounts from using a non whitelisted account's
+/// funds.
+/// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked
+/// instead. Thus the intermediary should only deposit and mint using assets initially owned by its current initiator.
 /// @dev No-ops are allowed.
 /// @dev Zero checks are not systematically performed.
 contract WhitelistSendAssetsGate is IWhitelistSendAssetsGate {
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -12,7 +12,7 @@ import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
 /// @dev The whitelisted accounts are responsible for the assets they deposit to the vault.
 /// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead.
 /// @dev The intermediary should only deposit assets when forwarding them from the initiator.
-///@dev No-ops are allowed.
+/// @dev No-ops are allowed.
 /// @dev Zero checks are not systematically performed.
 contract WhitelistSendAssetsGate is IWhitelistSendAssetsGate {
     address public whitelister;
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -9,6 +9,7 @@ import {
 } from "./interfaces/IWhitelistSendAssetsGate.sol";
 import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
 
+/// @dev Using this gate allows to restrict who the funds are initially owned by in a vault's deposits/mints.
 /// @dev The whitelisted accounts are responsible for the assets they deposit to the vault.
 /// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead.
 /// @dev The intermediary should only deposit and mint using assets initially owned by its current initiator.
--- a/src/periphery/WhitelistSendAssetsGate.sol
+++ b/src/periphery/WhitelistSendAssetsGate.sol
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.28;
+
+import {
+    IWhitelistSendAssetsGate,
+    IIntermediary,
+    SET_IS_WHITELISTED_TYPEHASH
+} from "./interfaces/IWhitelistSendAssetsGate.sol";
+import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
+
+/// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead.
+contract WhitelistSendAssetsGate is IWhitelistSendAssetsGate {
+    address public whitelister;
+    mapping(address => uint256) public nonces;
+    mapping(address => bool) public isIntermediary;
+    mapping(address => bool) public isWhitelisted;
+
+    constructor(address _whitelister) {
+        whitelister = _whitelister;
+        emit Constructor(_whitelister);
+    }
+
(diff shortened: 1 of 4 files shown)


src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    function setWhitelister(address newWhitelister) external {
+        require(msg.sender == whitelister, NotWhitelister());
+        whitelister = newWhitelister;
+        emit SetWhitelister(newWhitelister);
+    }
+
+    function setIsWhitelisted(address account, bool newIsWhitelisted) external {
+        require(msg.sender == whitelister, NotWhitelister());
+        isWhitelisted[account] = newIsWhitelisted;
```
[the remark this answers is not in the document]
MathisGD: that's not dumb, but that really really not classic... leaning towards not changing

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,83 @@
...
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, PermitDeadlineExpired());
+        bytes32 hashStruct =
+            keccak256(abi.encode(SET_IS_WHITELISTED_TYPEHASH, account, newIsWhitelisted, nonces[account]++, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), hashStruct));
+        address recovered = ecrecover(digest, v, r, s);
+        require(recovered != address(0) && recovered == whitelister, InvalidSigner());
+        isWhitelisted[account] = newIsWhitelisted;
+        emit SetIsWhitelisted(account, newIsWhitelisted);
```
MathisGD: should it have its own event?

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,83 @@
...
+            (bool success, bytes memory returnData) = address(this).delegatecall(data[i]);
+            if (!success) {
+                assembly ("memory-safe") {
+                    revert(add(32, returnData), mload(returnData))
+                }
+            }
+        }
+    }
+
+    function canSendAssets(address account) external view returns (bool) {
```
MathisGD, line 46:
```suggestion
    /// @dev Reverts if isIntermediary[account] but account reverts on initiator().
    function canSendAssets(address account) external view returns (bool) {
```

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,84 @@
...
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, PermitDeadlineExpired());
+        bytes32 hashStruct =
+            keccak256(abi.encode(SET_IS_WHITELISTED_TYPEHASH, account, newIsWhitelisted, nonces[account]++, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), hashStruct));
+        address recovered = ecrecover(digest, v, r, s);
+        require(recovered != address(0) && recovered == whitelister, InvalidSigner());
+        isWhitelisted[account] = newIsWhitelisted;
+        emit SetIsWhitelistedWithSig(account, newIsWhitelisted);
```
adhusson, line 85: we could log the signer here

src/periphery/interfaces/IWhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,48 @@
...
+}
+
+interface IWhitelistSendAssetsGate is ISendAssetsGate {
+    /* EVENTS */
+
+    event Constructor(address indexed whitelister);
+    event SetWhitelister(address indexed newWhitelister);
+    event SetIsWhitelisted(address indexed account, bool newIsWhitelisted);
+    event SetIsWhitelistedWithSig(address indexed account, bool newIsWhitelisted);
+    event SetIsIntermediary(address indexed intermediary, bool isIntermediary);
```
adhusson:
```suggestion
    event SetIsIntermediary(address indexed intermediary, bool newIsIntermediary);
```

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    function setIsIntermediary(address intermediary, bool newIsIntermediary) external {
+        require(msg.sender == whitelister, NotWhitelister());
+        isIntermediary[intermediary] = newIsIntermediary;
+        emit SetIsIntermediary(intermediary, newIsIntermediary);
+    }
+
+    /// @dev Signature malleability is not explicitly prevented but it is not a problem thanks to the nonce.
+    function setIsWhitelistedWithSig(
```
adhusson, line 70: What is the reason for adding a signature variant for whitelist and only for whitelist? So users call the whitelisting function with the whitelister data, and because there is not such as usecase for intermediary & whitelister change?

src/periphery/WhitelistSendAssetsGate.sol
MathisGD: it's always whitelister right?

src/periphery/WhitelistSendAssetsGate.sol
MathisGD: the idea is that the whitelister could let the user batch their own whitelisting with their deposit in the vault. that's pretty practical... without that the whitelister needs to do a transaction and once it landed the user can do their deposit.

I don't see such use-case for intermediary and whitelister

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,84 @@
...
+                assembly ("memory-safe") {
+                    revert(add(32, returnData), mload(returnData))
+                }
+            }
+        }
+    }
+
+    /// @dev Reverts if isIntermediary[account] but account reverts on initiator().
+    function canSendAssets(address account) external view returns (bool) {
+        return isWhitelisted[isIntermediary[account] ? IIntermediary(account).initiator() : account];
```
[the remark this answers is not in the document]
MathisGD: on purpose

src/periphery/WhitelistSendAssetsGate.sol
QGarchery, line 1: Wondering about the expected behavior in case of a removal from the whitelist (see [this question](https://morpholabs.slack.com/archives/C08A40KKN9W/p1781015512523409?thread_ts=1780597734.423279&cid=C08A40KKN9W)) ? Maybe not allowing removal makes more sense ?

QGarchery, line 1: Should we mention that the vault that have this gate cannot be easily used as collateral ?

QGarchery (commented):
See also https://github.com/morpho-org/vault-v2/pull/929

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.28;
```
QGarchery, line 3: Not using the latest version is fine because we are not impacted by those issues:
- [transient storage clearing](https://www.soliditylang.org/blog/2026/02/18/transient-storage-clearing-helper-collision-bug/)
- [lost array write](https://www.soliditylang.org/blog/2025/12/18/lost-storage-array-write-on-slot-overflow-bug/)

src/periphery/WhitelistSendAssetsGate.sol
MathisGD: Why so?

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,84 @@
...
+        address account,
+        bool newIsWhitelisted,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, PermitDeadlineExpired());
+        bytes32 hashStruct =
+            keccak256(abi.encode(SET_IS_WHITELISTED_TYPEHASH, account, newIsWhitelisted, nonces[account]++, deadline));
```
adhusson, line 80: nonces are keyed to the whitelisted account, which is unusual (for instance in permit nonces are not keyed to the spender)

MathisGD: yes, and they are also not keyed to the signer (which is also unusual).

I think that the current setup is ok though

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,91 @@
...
+    IIntermediary,
+    SET_IS_WHITELISTED_TYPEHASH
+} from "./interfaces/IWhitelistSendAssetsGate.sol";
+import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
+
+/// @dev Using this gate allows to restrict who the funds are initially owned by in a vault's deposits/mints.
+/// @dev As with any send assets gate, nothing prevents whitelisted accounts from using a non whitelisted account's
+/// funds.
+/// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked
+/// instead. Thus the intermediary should only deposit and mint using assets initially owned by its current initiator.
```
adhusson: "instead of" what is not clear at this stage.

```suggestion
/// @dev If account is registered as a trusted intermediary, IIntermediary(account).initiator() is checked for whitelisted status
/// instead of account itself. Thus the intermediary should only deposit and mint using assets initially owned by its current initiator.
```

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    function setIsIntermediary(address intermediary, bool newIsIntermediary) external {
+        require(msg.sender == whitelister, NotWhitelister());
+        isIntermediary[intermediary] = newIsIntermediary;
+        emit SetIsIntermediary(intermediary, newIsIntermediary);
+    }
+
+    /// @dev Signature malleability is not explicitly prevented but it is not a problem thanks to the nonce.
+    /// @dev Allows to batch setIsWhitelisted with the deposit, without requiring a transaction from the whitelister.
```
adhusson, line 69:
```suggestion
    /// @dev Allows depositors to batch setIsWhitelisted with the deposit, without requiring a transaction from the whitelister.
```

QGarchery (approved):

adhusson (approved):

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -0,0 +1,91 @@
...
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external {
+        require(deadline >= block.timestamp, DeadlineExpired());
+        bytes32 hashStruct =
+            keccak256(abi.encode(SET_IS_WHITELISTED_TYPEHASH, account, newIsWhitelisted, nonces[account]++, deadline));
+        bytes32 digest = keccak256(abi.encodePacked("\x19\x01", DOMAIN_SEPARATOR(), hashStruct));
+        address recovered = ecrecover(digest, v, r, s);
+        require(recovered != address(0) && recovered == whitelister, InvalidSigner());
```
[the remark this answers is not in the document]
MathisGD: we don't want to support it but relevant

PR #929, opened 2026-06-09
Review whitelist

QGarchery:


src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -9,12 +9,16 @@ import {
 } from "./interfaces/IWhitelistSendAssetsGate.sol";
 import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
 
+/// @dev The whitelisted accounts are responsible for the assets they deposit to the vault.
```
MathisGD: What does this mean?

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -9,12 +9,16 @@ import {
...
 /// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead.
+/// @dev The intermediary should only deposit assets when forwarding them from the initiator.
```
adhusson:
```suggestion
/// @dev The intermediary should only send assets to the vault when forwarding them from the initiator.
```

adhusson (approved):

src/periphery/WhitelistSendAssetsGate.sol
MathisGD:
```suggestion
/// @dev intermediaries should only deposit assets on behalf of the initiator.
```

MathisGD: should we say what is the "initiator"

QGarchery: @adhusson your suggestion can be confused with assets transfer. Maybe we can say `deposit/mint` instead of just `deposit`. wdyt ?

@MathisGD your suggestion misses an important point: even if you only deposit on behalf of the initiator, it could be that the assets come from another party. It should really pull assets from the initiator to be consistent with what the gate checks for the whitelisted path (not the intermediary path)

MathisGD:
```suggestion
/// @dev Thus intermediaries must only pull assets from the initiator.
```

src/periphery/WhitelistSendAssetsGate.sol
MathisGD:
```suggestion
/// @dev As it is a send assets gate, whitelisted accounts can deposit on behalf of non whitelisted accounts.
```

src/periphery/WhitelistSendAssetsGate.sol
adhusson: > `deposit/mint`

good for me but the sentence doesn't work then (`should only deposit/mint assets...`). 

We can say

```suggestion
/// @dev The intermediary should only deposit and mint using assets initially owned by its current initiator.
```

> and pull assets from the initiator.

It's going into implementation a little bit, the initiator may do a flow where it first sends the assets to the intermediary.

src/periphery/WhitelistSendAssetsGate.sol
QGarchery: Thinking of something like : Authorized accounts (either directly whitelisted or indirectly through the intermediary) can deposit on behalf of non whitelisted accounts.

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -9,12 +9,16 @@ import {
 } from "./interfaces/IWhitelistSendAssetsGate.sol";
 import {DOMAIN_TYPEHASH} from "../libraries/ConstantsLib.sol";
```
MathisGD, line 11:
```suggestion

/// @dev Using this gate allows to restrict who the funds are initially owned by in a vault's deposits/mints.
```

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -9,12 +9,17 @@ import {
...
 
+/// @dev Using this gate allows to restrict who the funds are initially owned by in a vault's deposits/mints.
+/// @dev As it is a send assets gate, whitelisted accounts can deposit on behalf of non whitelisted accounts.
```
adhusson:
```suggestion
/// @dev As with any send assets gate, nothing prevents whitelisted accounts from using a non whitelisted account's funds.
```

src/periphery/WhitelistSendAssetsGate.sol
```diff
@@ -9,12 +9,17 @@ import {
...
 /// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead.
+/// @dev The intermediary should only deposit and mint using assets initially owned by its current initiator.
```
MathisGD: it follows
```suggestion
/// @dev If `account` is registered as a trusted intermediary, IIntermediary(account).initiator() is checked instead. Thus the intermediary should only deposit and mint using assets initially owned by its current initiator.
```

PR #933, opened 2026-06-15
Update documentation for receiveSharesGate

MathisGD:
Clarify reverting gates.

### DIFF
--- a/src/VaultV2.sol
+++ b/src/VaultV2.sol
@@ -153,7 +153,7 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 /// @dev receiveSharesGate:
 ///     - Gates receiving shares.
 ///     - Can lock users out of getting back their shares deposited on an other contract.
-///     - Can DOS the vault if it reverts.
+///     - Can block all vault interactions if it reverts or consumes too much gas.
 /// @dev sendSharesGate:
 ///     - Gates sending shares.
 ///     - Can lock users out of exiting the vault.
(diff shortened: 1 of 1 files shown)


src/VaultV2.sol
```diff
@@ -150,10 +150,10 @@ import {IReceiveSharesGate, ISendSharesGate, IReceiveAssetsGate, ISendAssetsGate
 ///
 /// GATES
 /// @dev Set to 0 to disable a gate.
-/// @dev Gates must never revert, nor consume too much gas.
 /// @dev receiveSharesGate:
 ///     - Gates receiving shares.
 ///     - Can lock users out of getting back their shares deposited on an other contract.
+///     - Can DOS the vault if it reverts.
```
QGarchery:
```suggestion
///     - Can DOS the vault if it reverts or consumes too much gas.
```

qdrn:
```suggestion
///     - Can DoS the vault if it reverts or consumes too much gas.
```

qdrn:
```suggestion
///     - Can block all vault interactions if it reverts or consumes too much gas.
```
More verbose but easier to read.

adhusson: With a bad gate alone, it's not easy to DoS by gas consumption. The max gas given to a call is 63/64th of remaining gas, so with a ~16M gas limit per tx that leaves 250k gas after the call. It cannot block shares transfer. With two > 0 fees & two fee recipients, it can block accruals (but they can be set to 0 / disabled even if the gate is abdicated). It may block deposits if asset transfer or liquidity adapter allocation consumes too much gas.

QGarchery (approved):

qdrn (approved):

adhusson (approved):

PR #937, opened 2026-06-16
public allocator

MathisGD:
main difference with vault-v1 PA:
- boolean flow out instead of granular because it's enough control, and they seem in practice to be most of the time either fully closed or fully open.
- allocation cap (same as the absolute cap in vault, but set by allocators and without timelocks) the instead of flow-in cap, because it seems to align better to what curators are trying to control.
- inherits directly from the vault roles: allocators can set/unset flags; curator can set the ETH fee; and sentinel can set canDeallocate and unset canAllocate.


src/periphery/blue-public-allocator/BluePublicAllocator.sol
adhusson, line 1: > entry controls but no exits control because anyway the liquidity could have been borrowed

not clear to me, what if there are market A B C and allocator is ok with B<->C but also wants liquidity to be used in market A?

MathisGD: sorry I didn't get your point

adhusson: If there is liquidity in market A, 2 possibilities for it to get out are 1) it can get out by being borrowed and 2) it can get out by being put in another market through a public allocator. 

The allocator may want some of his liquidity to be borrowed in market A in particular / he may not be indifferent between "liquidity goes away because of public allocator" and "liquidity goes away because it got borrowed".

MathisGD: right, and i think that can be useful to lower the rate on a market. will add canDeallocate bool

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,76 @@
...
+        address deallocateAdapter,
+        bytes calldata deallocateData,
+        address allocateAdapter,
+        bytes calldata allocateData,
+        uint128 assets
+    ) external payable {
+        require(msg.value == fee[vault], IncorrectFee());
+        if (msg.value > 0) {
+            (bool success,) = payable(IVaultV2(vault).curator()).call{value: msg.value}("");
+            require(success, FeeTransferFailed());
```
MathisGD: not sure it should revert on fail actually (if the curator is not payable).

but then there is the try-catch gas manipulation issue? Maybe not a thing here with a simple transfer.

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,76 @@
...
+        emit SetCanAllocate(msg.sender, vault, adapter, data, newCanAllocate);
+    }
+
+    function setCanDeallocate(address vault, address adapter, bytes calldata data, bool newCanDeallocate) external {
+        require(
+            IVaultV2(vault).isAllocator(msg.sender) || (!newCanDeallocate && IVaultV2(vault).isSentinel(msg.sender)),
+            Unauthorized()
+        );
+        canDeallocate[vault][keccak256(abi.encode(adapter, data))] = newCanDeallocate;
+        emit SetCanDeallocate(msg.sender, vault, adapter, data, newCanDeallocate);
```
MathisGD: should we log the key only?

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,76 @@
...
+    function setFee(address vault, uint256 newFee) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
+        fee[vault] = newFee;
+        emit SetFee(msg.sender, vault, newFee);
+    }
+
+    /* PUBLIC FUNCTION */
+
+    /// @dev The vault's caps are still enforced on the allocation, so this call reverts if it would exceed them.
+    function reallocate(
```
MathisGD, line 98: should we add multicall or loop in reallocate or the bundle will handle it

src/periphery/PublicAllocator.sol
MathisGD: Actually it's a bad idea

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,76 @@
...
+        bytes32 deallocateKey = keccak256(abi.encode(deallocateAdapter, deallocateData));
+        require(canDeallocate[vault][deallocateKey], CannotDeallocate());
+
+        bytes32 allocateKey = keccak256(abi.encode(allocateAdapter, allocateData));
+        require(canAllocate[vault][allocateKey], CannotAllocate());
+
+        IVaultV2(vault).deallocate(deallocateAdapter, deallocateData, assets);
+        IVaultV2(vault).allocate(allocateAdapter, allocateData, assets);
+
+        emit Reallocate(msg.sender, vault, allocateKey, deallocateKey, assets);
```
MathisGD: should we log the full data?

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IPublicAllocator} from "./interfaces/IPublicAllocator.sol";
+
+/// @dev To be usable, the PublicAllocator must be set as an allocator of the vault.
+/// @dev The PublicAllocator inherits the vault's roles. The vault's allocators can enable and disable canAllocate and
+/// canDeallocate; the vault's sentinels can disable canAllocate and enable canDeallocate, to cut off public inflows and
+/// allow public outflows for derisking; the vault's curator sets the ETH penalty.
+/// @dev Each reallocate call costs an ETH penalty in native currency, set per vault by the curator. The ETH penalty is
+/// accrued per vault and can be claimed by the vault's curator.
+/// @dev No-ops are allowed. Zero checks are not systematically performed.
```
MathisGD:
```suggestion
/// @dev No-ops are allowed. Zero checks are not performed.
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.28;
+
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IPublicAllocator} from "./interfaces/IPublicAllocator.sol";
+
+/// @dev To be usable, the PublicAllocator must be set as an allocator of the vault.
+/// @dev The PublicAllocator inherits the vault's roles. The vault's allocators can enable and disable canAllocate and
+/// canDeallocate; the vault's sentinels can disable canAllocate and enable canDeallocate, to cut off public inflows and
+/// allow public outflows for derisking; the vault's curator sets the ETH penalty.
```
MathisGD:
```suggestion
/// allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IPublicAllocator} from "./interfaces/IPublicAllocator.sol";
+
+/// @dev To be usable, the PublicAllocator must be set as an allocator of the vault.
+/// @dev The PublicAllocator inherits the vault's roles. The vault's allocators can enable and disable canAllocate and
+/// canDeallocate; the vault's sentinels can disable canAllocate and enable canDeallocate, to cut off public inflows and
+/// allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
+/// @dev Each reallocate call costs an ETH penalty in native currency, set per vault by the curator. The ETH penalty is
+/// accrued per vault and can be claimed by the vault's curator.
```
MathisGD:
```suggestion
/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is
/// accrued per vault and can be claimed by the vault's curator.
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IPublicAllocator} from "./interfaces/IPublicAllocator.sol";
+
+/// @dev To be usable, the PublicAllocator must be set as an allocator of the vault.
+/// @dev The PublicAllocator inherits the vault's roles. The vault's allocators can enable and disable canAllocate and
+/// canDeallocate; the vault's sentinels can disable canAllocate and enable canDeallocate, to cut off public inflows and
+/// allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is
+/// accrued per vault and can be claimed by the vault's curator.
```
MathisGD:
```suggestion
/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued per
/// vault and can be claimed by the vault's curator.
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+/// @dev No-ops are allowed. Zero checks are not performed.
+contract PublicAllocator is IPublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 key => bool)) public canAllocate;
+    mapping(address vault => mapping(bytes32 key => bool)) public canDeallocate;
+    mapping(address vault => uint256) public ethPenalty;
+    mapping(address vault => uint256) public accruedEthPenalty;
+
+    /* CONFIGURATION FUNCTIONS */
```
MathisGD:
```suggestion
    /* AUTHORIZED FUNCTIONS */
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    function setEthPenalty(address vault, uint256 newEthPenalty) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
+        ethPenalty[vault] = newEthPenalty;
+        emit SetEthPenalty(msg.sender, vault, newEthPenalty);
+    }
+
+    /* CLAIM FUNCTION */
+
```
MathisGD:
```suggestion
```

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,85 @@
...
+/// canDeallocate; the vault's sentinels can disable canAllocate and enable canDeallocate, to cut off public inflows and
+/// allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued per
+/// vault and can be claimed by the vault's curator.
+/// @dev No-ops are allowed. Zero checks are not performed.
+contract PublicAllocator is IPublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 key => bool)) public canAllocate;
+    mapping(address vault => mapping(bytes32 key => bool)) public canDeallocate;
```
MathisGD: not worth packing because they are never read together for the same key

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD, line 1: to have in mind: https://github.com/morpho-org/vault-v2/blob/549b113106e505b83952c7b296622534338d5f92/src/VaultV2.sol#L66

src/periphery/PublicAllocator.sol
peyha: probably not worth it imo
we can already re-construct the PA state with the current log and we could use `Allocate` and `Deallocate` events from vault-v2 when msg.sender is the PA to get adapter level PA data. It would still be difficult to link the address calling `reallocate` to this adapter level data but I don't see it as a big deal

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,85 @@
...
+/// allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued
+/// per vault and can be claimed by the vault's curator.
+/// @dev No-ops are allowed. Zero checks are not performed.
+contract PublicAllocator is IPublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 key => bool)) public canAllocate;
+    mapping(address vault => mapping(bytes32 key => bool)) public canDeallocate;
+    mapping(address vault => uint256) public ethPenalty;
```
peyha:
```suggestion
    mapping(address vault => uint256) public nativeTokenPenalty;
```
in case the gas token is anything other than eth (e.g. tempo, arc, merlin chain, ..)

MathisGD: not mandatory imo, there is already "ether" in solidity. and it reads better

src/periphery/interfaces/IPublicAllocator.sol
```diff
@@ -0,0 +1,63 @@
...
+    function claimEthPenalty(address vault, address payable receiver) external;
+    function reallocate(
+        address vault,
+        address deallocateAdapter,
+        MarketParams calldata deallocateMarketParams,
+        address allocateAdapter,
+        MarketParams calldata allocateMarketParams,
+        uint128 assets
+    ) external payable;
+    /// @dev Allocates assets from the vault's idle balance.
```
MathisGD:
```suggestion
```

src/periphery/interfaces/IPublicAllocator.sol
```diff
@@ -0,0 +1,63 @@
...
+    error Unauthorized();
+    error AbsoluteCapExceeded();
+    error CannotDeallocate();
+    error EthTransferFailed();
+    error IncorrectEthPenalty();
+
+    /* VIEW */
+
+    /// @dev A Morpho Market V1 market of a vault is keyed by its per-market vault id,
+    /// id = keccak256(abi.encode("this/marketParams", adapter, marketParams)), exactly as in the vault's caps.
```
MathisGD:
```suggestion
```

src/periphery/interfaces/IPublicAllocator.sol
```diff
@@ -0,0 +1,60 @@
...
+        address indexed sender, address indexed vault, address adapter, MarketParams marketParams, uint256 absoluteCap
+    );
+    event SetCanDeallocate(
+        address indexed sender, address indexed vault, address adapter, MarketParams marketParams, bool canDeallocate
+    );
+    event SetCanDeallocateFromIdle(address indexed sender, address indexed vault, bool canDeallocate);
+    event SetEthPenalty(address indexed sender, address indexed vault, uint256 newEthPenalty);
+    event ClaimEthPenalty(address indexed sender, address indexed vault, uint256 claimed, address receiver);
+    event Reallocate(
+        address indexed sender, address indexed vault, bytes32 indexed allocateId, bytes32 deallocateId, uint128 assets
```
peyha:
```suggestion
        address sender, address indexed vault, bytes32 indexed allocateId, bytes32 indexed deallocateId, uint128 assets
```
not sure what would be best but it feels weird to have `allocateId` indexed and not `deallocateId`

src/periphery/blue-public-allocator/BluePublicAllocator.sol
peyha: from [this dune query](https://dune.com/queries/7998218?category=third_party_data&contract=PublicAllocator), ~10% of blue's reallocation have deallocated >= 2 markets so reallocating from multiple market is significant
not sure if people would end up reallocating by deallocating by multiple blue adapter or from different adapter though

MathisGD: I think that it should be handled in the borrow bundle

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+        IVaultV2(vault).deallocate(deallocateAdapter, abi.encode(deallocateMarketParams), assets);
+        IVaultV2(vault).allocate(allocateAdapter, abi.encode(allocateMarketParams), assets);
+
+        bytes32 allocateId = marketId(allocateAdapter, allocateMarketParams);
+        require(IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId], AbsoluteCapExceeded());
+
+        emit Reallocate(msg.sender, vault, allocateId, deallocateId, assets);
+    }
+
+    function allocate(address vault, address adapter, MarketParams calldata marketParams, uint128 assets)
```
MathisGD: could be merged with reallocate, not sure

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+    function reallocate(
+        address vault,
+        address deallocateAdapter,
+        MarketParams calldata deallocateMarketParams,
+        address allocateAdapter,
+        MarketParams calldata allocateMarketParams,
+        uint128 assets
+    ) external payable {
+        require(msg.value == ethPenalty[vault], IncorrectEthPenalty());
+        if (msg.value > 0) accruedEthPenalty[vault] += msg.value;
```
peyha: there should probably be an event for this, otherwise we cannot re-construct the state from event ([cf this](https://morpholabs.slack.com/archives/C02N7CZ088N/p1715087292468609))

peyha: note that with [eip-7708](https://forkcast.org/eips/7708/) might solve this on glamsterdam

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+    mapping(address vault => uint256) public ethPenalty;
+    mapping(address vault => uint256) public accruedEthPenalty;
+
+    /* AUTHORIZED FUNCTIONS */
+
+    function setAbsoluteCap(address vault, address adapter, MarketParams calldata marketParams, uint256 newAbsoluteCap)
+        external
+    {
+        bytes32 id = marketId(adapter, marketParams);
+        require(
```
peyha: should we check that `IVaultV2(vault).isAdapter(adapter)` ?

peyha: if we don't do it, an allocator could maliciously or mistakenly set absolute caps on non existing allocators. does not seem like a problem and it costs one cold sload

src/periphery/blue-public-allocator/BluePublicAllocator.sol
peyha: why ? using a an array of deallocation data here would remove some complexity in the borrow bundle, simplify the code by merging `allocate` with `reallocate` and be more gas efficient for multiple deallocation reallocate

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.28;
```
peyha, line 3:
```suggestion
pragma solidity 0.8.36;
```
because some bug were fixed ?

src/periphery/BlueAdapterV2PublicAllocator.sol
MathisGD: yes I don't think that it's a problem. similarly in vault-v2 nothing is checked about the ids for which one might increase caps

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+            IVaultV2(vault).isAllocator(msg.sender) || (newCanDeallocate && IVaultV2(vault).isSentinel(msg.sender)),
+            Unauthorized()
+        );
+        canDeallocate[vault][marketId(adapter, marketParams)] = newCanDeallocate;
+        emit SetCanDeallocate(msg.sender, vault, adapter, marketParams, newCanDeallocate);
+    }
+
+    function setCanDeallocateFromIdle(address vault, bool newCanDeallocate) external {
+        require(
+            IVaultV2(vault).isAllocator(msg.sender) || (newCanDeallocate && IVaultV2(vault).isSentinel(msg.sender)),
```
adhusson: should a sentinel be able to do this? Shouldn't it be `|| (!newCanDeallocate && ...)`.

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+        require(msg.value == ethPenalty[vault], IncorrectEthPenalty());
+        if (msg.value > 0) accruedEthPenalty[vault] += msg.value;
+        bytes32 deallocateId = marketId(deallocateAdapter, deallocateMarketParams);
+        require(canDeallocate[vault][deallocateId], CannotDeallocate());
+
+        IVaultV2(vault).deallocate(deallocateAdapter, abi.encode(deallocateMarketParams), assets);
+        IVaultV2(vault).allocate(allocateAdapter, abi.encode(allocateMarketParams), assets);
+
+        bytes32 allocateId = marketId(allocateAdapter, allocateMarketParams);
+        require(IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId], AbsoluteCapExceeded());
```
peyha, line 118: would it make sense to only rely on the vault cap ? (and to use a bool here)

src/periphery/PublicAllocator.sol
MathisGD: my thinking was: it can only derisk to enable the liquidity to get out of a given market, so yes they should be able to. and they shouldn't be able to deactivate because it technically increases the risk

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: the goal is really to have a second layer of caps, that is 
- set by the allocator (not by heavy multisig)
- and without timelock

src/periphery/PublicAllocator.sol
adhusson: but here it's `setCanDeallocateFromIdle` (not `setCanDeallocate`) so the risk is reversed

adhusson: maybe with a name like `setCanAllocate` it would be easier

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+        bytes32 allocateId = marketId(adapter, marketParams);
+        require(IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId], AbsoluteCapExceeded());
+
+        emit Allocate(msg.sender, vault, allocateId, assets, msg.value);
+    }
+
+    /* INTERNAL */
+
+    /// @dev Returns the market's per-market vault id, exactly as keyed by the MorphoMarketV1AdapterV2.
+    function marketId(address adapter, MarketParams calldata marketParams) internal pure returns (bytes32) {
```
adhusson: in the context of midnight what about a less ambiguous name

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+        accruedEthPenalty[vault] = 0;
+        (bool success,) = receiver.call{value: claimed}("");
+        require(success, EthTransferFailed());
+
+        emit ClaimEthPenalty(msg.sender, vault, claimed, receiver);
+    }
+
+    /* PUBLIC FUNCTION */
+
+    function reallocate(
```
adhusson, line 98: (Fable find btw) 

Imagine `reallocate` is called with an `allocateAdapter` that has `absoluteCap[vault][marketId(allocateAdapter,data)] == 0` for all `data`. And `allocateMarketParams` is set to data accepted by `allocateAdapter`, but it doesn't match any real `MarketParams` (as long as `allocateAdapter` accepts data <= the size of a MarketParams struct it works). That adapter will not return `allocateId` in its list. So the allocation of `allocateId` in the vault will not increase. So the check `IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId]` will succeed.

adhusson: since this is for the Blue adapter only we can just check the factory

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,125 @@
...
+/// inflows and allow public outflows for derisking; the vault's curator sets and claims the ETH penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued
+/// per vault and can be claimed by the vault's curator.
+/// @dev The vault's caps are still enforced on the allocation, so this call reverts if it would exceed them.
+/// @dev No-ops are allowed. Zero checks are not performed.
+contract PublicAllocator is IPublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 id => uint256)) public absoluteCap;
+    mapping(address vault => mapping(bytes32 id => bool)) public canDeallocate;
```
adhusson, line 27: Would it make sense to have an absoluteCap and an absoluteFloor? It's symmetric but don't know how useful. You would have canDeallocate <=> floor=0.

src/periphery/PublicAllocator.sol
MathisGD: Ah yes mb. Will change

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: Im not sure that it has any use. And it makes the borrower algo a bit more complex. Worth it?

adhusson: Not sure either way, but also it would need to store `uint.max - actualFloor` so the default value is right.

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: wait it relies on having enabled an adapter that doesn't return any id? it's obviously a bad setup (already documented) no?

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: I don't feel it's worth it

src/periphery/PublicAllocator.sol
MathisGD: yes id is pretty overloaded... not sure what to put though

src/periphery/blue-public-allocator/BluePublicAllocator.sol
adhusson: The enabled adapter can return an id, but a different one. It will increase its allocation in that id, and not in the id the public allocator looks at.

MathisGD: oh yes ok. tricky one. the weird thing is that we really can't access the ids that were returned by the adapter... so I guess yes the only solution is to check the factory

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,126 @@
...
+        emit SetCanDeallocateFromIdle(msg.sender, vault, newCanDeallocate);
+    }
+
+    function setNativePenalty(address vault, uint256 newNativePenalty) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
+        nativePenalty[vault] = newNativePenalty;
+        emit SetNativePenalty(msg.sender, vault, newNativePenalty);
+    }
+
+    function claimNativePenalty(address vault, address payable receiver) external {
```
peyha, line 87: another way to have this would be to implement a `recipient` set by the curator that would receive the native penalty 
This way the curator would have to interact less with the contract

src/periphery/PublicAllocator.sol
peyha: `marketLevelAdapterId` ? not sure we can have something short and unambiguous

src/periphery/blue-public-allocator/interfaces/IBluePublicAllocator.sol
peyha, line 1: should we provide a very general `IPublicAllocator.sol` in case other people want to build public allocators for their own adapters ?
we could also rename this one `BluePublicAllocator` but not sure it's a good idea because we only have this public allocator planned atm

src/periphery/PublicAllocator.sol
peyha: probably with a if else + special idle case `deallocateAdapter=address(0)` but it does not look so good

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,126 @@
...
+        IVaultV2(vault).deallocate(deallocateAdapter, abi.encode(deallocateMarketParams), assets);
+        IVaultV2(vault).allocate(allocateAdapter, abi.encode(allocateMarketParams), assets);
+
+        bytes32 allocateId = marketId(allocateAdapter, allocateMarketParams);
+        require(IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId], AbsoluteCapExceeded());
+
+        emit Reallocate(msg.sender, vault, allocateId, deallocateId, assets, msg.value);
+    }
+
+    function allocate(address vault, address adapter, MarketParams calldata marketParams, uint128 assets)
```
peyha:
```suggestion
    function allocateFromIdle(address vault, address adapter, MarketParams calldata marketParams, uint128 assets)
```
a bit clearer

test/periphery/PublicAllocatorTest.sol
```diff
@@ -0,0 +1,486 @@
...
+pragma solidity ^0.8.28;
+
+import "../integration/MorphoMarketV1IntegrationTest.sol";
+
+import {PublicAllocator} from "../../src/periphery/PublicAllocator.sol";
+import {IPublicAllocator} from "../../src/periphery/interfaces/IPublicAllocator.sol";
+
+contract RejectNative {}
+
+contract GasHungryNativeReceiver {
```
peyha:
```suggestion
contract DummyReceiver {
```
weird naming

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: I initially designed it this way. but then you have to manage the question of the gas spent by the receiver etc... didn't seem worth it

test/periphery/PublicAllocatorTest.sol
MathisGD: yes, actually these tests are path dependance, they are useless now

src/periphery/PublicAllocator.sol
MathisGD: agree

src/periphery/PublicAllocator.sol
MathisGD: let's go for `vaultBlueId`

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: actually let's add a comment about that in the public allocator contract

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: we should add in the contractor the adapter factory and then call it to check that the adapter is one we expect

test/periphery/PublicAllocatorTest.sol
MathisGD: we should remove them.

src/periphery/PublicAllocator.sol
MathisGD: chose not to, it's not that much more complexity for the bundles

src/periphery/blue-public-allocator/interfaces/IBluePublicAllocator.sol
MathisGD: let's just rename this contract `BlueAdapterV2PublicAllocator`. makes it clear that it's tied to this adapter

src/periphery/PublicAllocator.sol
```diff
@@ -25,6 +30,12 @@ contract PublicAllocator is IPublicAllocator {
     mapping(address vault => uint256) public nativePenalty;
     mapping(address vault => uint256) public accruedNativePenalty;
```
MathisGD: let's pack these 3 values: bool, uint120, uint120

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -120,7 +131,9 @@ contract PublicAllocator is IPublicAllocator {
     /* INTERNAL */
 
     /// @dev Returns the market's per-market vault id, exactly as keyed by the MorphoMarketV1AdapterV2.
-    function vaultBlueId(address adapter, MarketParams calldata marketParams) internal pure returns (bytes32) {
+    /// @dev Reverts if the adapter was not created by the expected factory, restricting all paths to Blue adapters.
+    function vaultBlueId(address adapter, MarketParams calldata marketParams) internal view returns (bytes32) {
+        require(IMorphoMarketV1AdapterV2Factory(adapterFactory).isMorphoMarketV1AdapterV2(adapter), NotBlueAdapter());
```
MathisGD: don't do the check here, do it in the top level functions at the top

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,175 @@
...
+        address deallocateAdapter,
+        MarketParams calldata deallocateMarketParams,
+        address allocateAdapter,
+        MarketParams calldata allocateMarketParams,
+        uint128 assets
+    ) external payable {
+        require(
+            IMorphoMarketV1AdapterV2Factory(adapterFactory).isMorphoMarketV1AdapterV2(deallocateAdapter),
+            NotBlueAdapter()
+        );
```
MathisGD: technically we don't need to check that, should we do it anyway?

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,175 @@
...
+        require(success, NativeTransferFailed());
+
+        emit ClaimNativePenalty(msg.sender, vault, claimed, receiver);
+    }
+
+    /* PUBLIC FUNCTION */
+
+    function reallocate(
+        address vault,
+        address deallocateAdapter,
```
MathisGD, line 100: do we want to support having two adapters actually?

src/periphery/blue-public-allocator/interfaces/IBluePublicAllocator.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity >=0.5.0;
+
+import {MarketParams} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+
```
MathisGD, line 6: we can deactivate fmt in this file to have each statement be 1 line

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: Claude, implement that please

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,168 @@
...
+        require(success, NativeTransferFailed());
+
+        emit ClaimNativePenalty(msg.sender, vault, claimed, receiver);
+    }
+
+    /* PUBLIC FUNCTION */
+
+    function reallocate(
+        address vault,
+        address adapter,
```
MathisGD: single adapter now, tell me if you think that's not ok

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,169 @@
...
+/// @dev Specialized to Morpho Blue allocations through the MorphoMarketV1AdapterV2.
+/// @dev To be usable, the BlueAdapterV2PublicAllocator must be set as an allocator of the vault.
+/// @dev The BlueAdapterV2PublicAllocator inherits the vault's roles. The vault's allocators can set the absolute cap
+/// and canDeallocate and canDeallocateFromIdle; the vault's sentinels can decrease the absolute cap, enable
+/// canDeallocate, and disable canDeallocateFromIdle, to cut off public inflows and allow public outflows for derisking;
+/// the vault's curator sets and claims the native penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued
+/// per vault and can be claimed by the vault's curator.
+/// @dev The vault's caps are still enforced on the allocation, so this call reverts if it would exceed them.
+/// @dev Relative caps can still be manipulated by allocators through short-term deposits, but it requires capital.
```
MathisGD:
```suggestion
/// @dev The Public Allocator opens the door for anybody to manipulate relative caps through short-term deposits (but it requires capital).
```

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -67,21 +91,23 @@ contract BlueAdapterV2PublicAllocator is IBlueAdapterV2PublicAllocator {
...
         );
-        canDeallocateFromIdle[vault] = newCanDeallocate;
+        _vaultData[vault].canDeallocateFromIdle = newCanDeallocate;
         emit SetCanDeallocateFromIdle(msg.sender, vault, newCanDeallocate);
     }
 
     function setNativePenalty(address vault, uint256 newNativePenalty) external {
         require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
-        nativePenalty[vault] = newNativePenalty;
+        require(newNativePenalty <= type(uint120).max, ErrorsLib.CastOverflow());
```
peyha: probably not worth importing `ErrorsLib` just for this error

src/periphery/PublicAllocator.sol
```diff
@@ -18,6 +19,15 @@ import {MarketParams} from "../../lib/morpho-blue/src/interfaces/IMorpho.sol";
 /// @dev The vault's caps are still enforced on the allocation, so this call reverts if it would exceed them.
 /// @dev No-ops are allowed. Zero checks are not performed.
 contract BlueAdapterV2PublicAllocator is IBlueAdapterV2PublicAllocator {
+    /* TYPES */
+
+    /// @dev Packed into a single storage slot: bool (1 byte) + uint120 (15 bytes) + uint120 (15 bytes) = 31 bytes.
+    struct VaultData {
+        bool canDeallocateFromIdle;
+        uint120 nativePenalty;
```
peyha: is it worth a comment to say that the native penalty cannot be accrued too much (log10(2^120)=36 so 18 digit with 18 decimals)

peyha (approved):

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+            IVaultV2(vault).isAllocator(msg.sender) || (!newCanDeallocate && IVaultV2(vault).isSentinel(msg.sender)),
+            Unauthorized()
+        );
+        _vaultData[vault].canDeallocateFromIdle = newCanDeallocate;
+        emit SetCanDeallocateFromIdle(msg.sender, vault, newCanDeallocate);
+    }
+
+    function setNativePenalty(address vault, uint256 newNativePenalty) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
+        require(newNativePenalty <= type(uint120).max, CastOverflow());
```
adhusson, line 81: should be tested

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+    function accruedNativePenalty(address vault) external view returns (uint256) {
+        return _vaultData[vault].accruedNativePenalty;
+    }
+
+    /* AUTHORIZED FUNCTIONS */
+
+    function setAbsoluteCap(address vault, address adapter, MarketParams calldata marketParams, uint256 newAbsoluteCap)
+        external
+    {
+        require(IMorphoMarketV1AdapterV2Factory(adapterFactory).isMorphoMarketV1AdapterV2(adapter), NotBlueAdapter());
```
adhusson: should be more tested (for all functions and not just with address zero but with any adapter that is just not in the factory)

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+import {IMorphoMarketV1AdapterV2Factory} from "../adapters/interfaces/IMorphoMarketV1AdapterV2Factory.sol";
+import {MarketParams} from "../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+
+/// @dev Specialized to Morpho Blue allocations through the MorphoMarketV1AdapterV2.
+/// @dev To be usable, the BlueAdapterV2PublicAllocator must be set as an allocator of the vault.
+/// @dev The BlueAdapterV2PublicAllocator inherits the vault's roles. The vault's allocators can set the absolute cap
+/// and canDeallocate and canDeallocateFromIdle; the vault's sentinels can decrease the absolute cap, enable
+/// canDeallocate, and disable canDeallocateFromIdle, to cut off public inflows and allow public outflows for derisking;
+/// the vault's curator sets and claims the native penalty.
+/// @dev Each reallocate call costs a penalty in native currency, set per vault by the curator. The penalty is accrued
```
adhusson:
```suggestion
/// @dev Each reallocate and allocateFromIdle call costs a penalty in native currency, set per vault by the curator. The penalty is accrued
```

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+/// per vault and can be claimed by the vault's curator.
+/// @dev The vault's caps are still enforced on the allocation, so this call reverts if it would exceed them.
```
adhusson: should we comment that if the liquidity adapter's market cannot be reallocated it's possible to make it impossible to withdraw the normal way by users

MathisGD: good remark. but it's pretty obvious no? same for idle

src/periphery/PublicAllocator.sol
QGarchery: It wouldn't be too costly to add the blue ids though right ? Actually maybe they could replace the vault ids, because they can be used to recompute the vault ids (you first to `idToMarketParams`)

src/periphery/interfaces/IBlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2026 Morpho Association
+pragma solidity >=0.5.0;
```
QGarchery:
```suggestion
pragma solidity >=0.8.0;
```
per our convention [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1782745338032559?thread_ts=1782740536.272999&cid=C08A40KKN9W)

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    /* IMMUTABLES */
+
+    address public immutable adapterFactory;
+
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 id => uint256)) public absoluteCap;
+    mapping(address vault => mapping(bytes32 id => bool)) public canDeallocate;
+    mapping(address vault => VaultData) internal _vaultData;
```
QGarchery:
```suggestion
    mapping(address vault => VaultData) public vaultData;
```
why not ?

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    function claimNativePenalty(address vault, address payable receiver) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
+
+        uint256 claimed = _vaultData[vault].accruedNativePenalty;
+        _vaultData[vault].accruedNativePenalty = 0;
+        (bool success,) = receiver.call{value: claimed}("");
+        require(success, NativeTransferFailed());
+
+        emit ClaimNativePenalty(msg.sender, vault, claimed, receiver);
```
QGarchery:
```suggestion
        emit ClaimNativePenalty(msg.sender, vault, claimed, receiver);
        (bool success,) = receiver.call{value: claimed}("");
        require(success, NativeTransferFailed());
```
to have nice ordering of events, since the call to the receiver could do other things

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+/// @dev No-ops are allowed. Zero checks are not performed.
+contract BlueAdapterV2PublicAllocator is IBlueAdapterV2PublicAllocator {
+    /* TYPES */
+
+    /// @dev Packed into a single storage slot: bool (1 byte) + uint120 (15 bytes) + uint120 (15 bytes) = 31 bytes.
+    struct VaultData {
+        bool canDeallocateFromIdle;
+        uint120 nativePenalty;
+        // uint120 max ~2^120 (~1e18 tokens at 18 decimals), so accrued native penalty cannot realistically overflow.
+        uint120 accruedNativePenalty;
```
QGarchery:
```suggestion
```
could we just do that and just sweep the whole native balance when claiming ? It’s possible that it’s too much, because you can force send native tokens (with the SELFDESTRUCT trick), but do we care ?

src/periphery/blue-public-allocator/BluePublicAllocator.sol
QGarchery, line 1: There is no mention that it should be used with VaultV2 (quite obvious though given that it is in vault-v2 repo). Maybe still worth it ? It begs the question because we have a factory check for the adapter but not for the vault

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+import {IVaultV2} from "../interfaces/IVaultV2.sol";
+import {IBlueAdapterV2PublicAllocator} from "./interfaces/IBlueAdapterV2PublicAllocator.sol";
+import {IMorphoMarketV1AdapterV2Factory} from "../adapters/interfaces/IMorphoMarketV1AdapterV2Factory.sol";
+import {MarketParams} from "../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+
+/// @dev Specialized to Morpho Blue allocations through the MorphoMarketV1AdapterV2.
+/// @dev To be usable, the BlueAdapterV2PublicAllocator must be set as an allocator of the vault.
+/// @dev The BlueAdapterV2PublicAllocator inherits the vault's roles. The vault's allocators can set the absolute cap
+/// and canDeallocate and canDeallocateFromIdle; the vault's sentinels can decrease the absolute cap, enable
```
QGarchery: ⚠️ There is an issue where sentinels are not able to effectively protect allocation, in this contract. This is because, even if they deallocate, an allocator could simply allocate again to the same market. 
Sentinels can't delist an allocator in the vaut, so the best they could do is to put the caps to zero on the vault. At that point the extra powers given to sentinels in this contract are not very useful

src/periphery/BlueAdapterV2PublicAllocator.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    function reallocate(
+        address vault,
+        address adapter,
+        MarketParams calldata deallocateMarketParams,
+        MarketParams calldata allocateMarketParams,
+        uint128 assets
+    ) external payable {
+        require(IMorphoMarketV1AdapterV2Factory(adapterFactory).isMorphoMarketV1AdapterV2(adapter), NotBlueAdapter());
+        require(msg.value == _vaultData[vault].nativePenalty, IncorrectNativePenalty());
```
QGarchery:
```suggestion
        require(msg.value >= _vaultData[vault].nativePenalty, IncorrectNativePenalty());
```
we could do that too. Use case would be anticipating a increase in fee. Probably not worth it though, just wanted to flag it, but feel free to close

src/periphery/blue-public-allocator/BluePublicAllocator.sol
QGarchery, line 1: We could add a README to periphery. It's a significant change of how vault-v2 can be used, that doesn't appear in the README at the root of the repository

QGarchery, line 1: I think we should add a comment saying that the allocate/deallocate functions can be made to revert if they are frontrun (by anyone, not just allocators). 
Allocate: someone allocates more than the cap of the vault minus the assets. 
Deallocate: someone removes enough so that the shares don’t cover assets.

QGarchery (commented):
See also #956

src/periphery/PublicAllocator.sol
MathisGD: well the public allocator is common to all vaults, so not really...

MathisGD: my best attempt at removing this accounting was by sending the ETH directly to the curator, but it creates a bunch of issues...

src/periphery/blue-public-allocator/BluePublicAllocator.sol
QGarchery, line 1: We don't have the property "if allocators don't do anything, then caps set by this contract are preserved (meaning if allocations are below the caps, then they stay below the caps)", which can be unintuitive so I think it should be documented. The reason why it's not true is because of the liquidity adapter, allocation can freely move above the cap set by this contract since it's independent

src/periphery/PublicAllocator.sol
QGarchery: ah yes of course 😅

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD: not that convinced, since we already have: allocation can move "by itself", without allocator intervention because of the liquidity market. and it's not documented.

src/periphery/BlueAdapterV2PublicAllocator.sol
MathisGD: at the end I reverted it

src/periphery/BlueAdapterV2PublicAllocator.sol
MathisGD: resolving optimistically

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD, line 1: recent changes:
- switch back to multi-adapter compatibility (for if we end-up whitelisting a new IRM, and it's cheap enough)
- general name, because technically it could work with future adapter
- addition of the activeAdapter mapping to fix the issue where you have an incompatible adapter enabled
- everything is managed by allocators

src/periphery/PublicAllocator.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    function setCanAllocateFromIdle(address vault, bool newCanDeallocate) external {
+        require(IVaultV2(vault).isAllocator(msg.sender), Unauthorized());
+        vaultData[vault].canAllocateFromIdle = newCanDeallocate;
+        emit SetCanAllocateFromIdle(msg.sender, vault, newCanDeallocate);
+    }
+
+    function setNativePenalty(address vault, uint256 newNativePenalty) external {
+        require(msg.sender == IVaultV2(vault).curator(), Unauthorized());
```
MathisGD: should it be doable by allocators?
- they can reallocate the way they want anyway
- simpler: PA is managed by allocators

MathisGD: discussed with @peyha, let's do it

src/periphery/PublicAllocator.sol
MathisGD: wait but we have them now, right?

src/periphery/blue-public-allocator/BluePublicAllocator.sol
MathisGD, line 1: should we create sub-folders in periphery

src/periphery/blue-public-allocator/interfaces/IBluePublicAllocator.sol
```diff
@@ -0,0 +1,58 @@
...
+    struct VaultData {
+        bool canAllocateFromIdle;
+        uint120 nativePenalty;
+        uint120 accruedNativePenalty;
+    }
+
+    /* EVENTS */
+
+    event SetAbsoluteCap(address indexed sender, address indexed vault, address adapter, MarketParams marketParams, uint256 absoluteCap);
```
peyha, line 18:
```suggestion
    event SetAbsoluteCap(address indexed sender, address indexed vault, address adapter, MarketParams indexed marketParams, uint256 absoluteCap);
```
in an ideal world we would do this but here solidity would hash marketParams so probably a bad idea

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+// Copyright (c) 2026 Morpho Association
+pragma solidity 0.8.28;
+
+import {IVaultV2} from "../../interfaces/IVaultV2.sol";
+import {IBluePublicAllocator} from "./interfaces/IBluePublicAllocator.sol";
+import {MarketParams} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol";
+
+/// @dev To be usable, the BluePublicAllocator must be set as an allocator of the vault.
+/// @dev Meant to be used with VaultV2 vaults only.
+/// @dev Active adapter must be MorphoMarketV1AdapterV2 adapters, otherwise the public allocator's absolute cap system
```
QGarchery:
```suggestion
/// @dev Active adapters must be MorphoMarketV1AdapterV2 adapters, otherwise the public allocator's absolute cap system
```

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+/// it requires capital).
+/// @dev reallocate and allocateFromIdle can be made to revert by anyone frontrunning them (not only allocators): an
+/// allocate reverts if the vault cap is filled first, a deallocate reverts if shares stop covering assets.
+contract BluePublicAllocator is IBluePublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 id => uint256)) public absoluteCap;
+    mapping(address vault => mapping(bytes32 id => bool)) public canDeallocate;
+    mapping(address vault => mapping(address adapter => bool)) public activeAdapter;
+    mapping(address vault => IBluePublicAllocator.VaultData) public vaultData;
```
QGarchery:
```suggestion
    mapping(address vault => VaultData) public vaultData;
```
and import it in the preamble

src/periphery/blue-public-allocator/BluePublicAllocator.sol
QGarchery: > since we already have: allocation can move "by itself", without allocator intervention because of the liquidity market. and it's not documented.

yes that's my whole point, this precise fact should be documented. But I see it is documented now

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /* AUTHORIZED FUNCTIONS */
+
+    function setActiveAdapter(address vault, address adapter, bool newActiveAdapter) external {
```
QGarchery: should we add a multicall ? Useful to batch allocators actions

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+/// it requires capital).
+/// @dev reallocate and allocateFromIdle can be made to revert by anyone frontrunning them (not only allocators): an
+/// allocate reverts if the vault cap is filled first, a deallocate reverts if shares stop covering assets.
+contract BluePublicAllocator is IBluePublicAllocator {
+    /* STORAGE */
+
+    mapping(address vault => mapping(bytes32 id => uint256)) public absoluteCap;
+    mapping(address vault => mapping(bytes32 id => bool)) public canDeallocate;
+    mapping(address vault => mapping(address adapter => bool)) public activeAdapter;
+    mapping(address vault => IBluePublicAllocator.VaultData) public vaultData;
```
peyha:
```suggestion
    mapping(address vault => VaultData) public vaultData;
```

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(canDeallocate[vault][deallocateId], CannotDeallocate());
+
+        IVaultV2(vault).deallocate(deallocateAdapter, abi.encode(deallocateMarketParams), assets);
+        IVaultV2(vault).allocate(allocateAdapter, abi.encode(allocateMarketParams), assets);
+
+        bytes32 allocateId = vaultBlueId(allocateAdapter, allocateMarketParams);
+        require(IVaultV2(vault).allocation(allocateId) <= absoluteCap[vault][allocateId], AbsoluteCapExceeded());
+
+        emit Reallocate(
+            msg.sender, vault, deallocateAdapter, allocateAdapter, allocateId, deallocateId, assets, msg.value
```
peyha:
```suggestion
            msg.sender, vault, deallocateAdapter, allocateAdapter, deallocateId, allocateId, assets, msg.value
```
nitpick

peyha: or 
```suggestion
            msg.sender, vault, deallocateAdapter, deallocateId,  allocateAdapter, allocateId, assets, msg.value
```

src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    function allocateFromIdle(address vault, address adapter, MarketParams calldata marketParams, uint128 assets)
+        external
+        payable
+    {
+        require(msg.value == vaultData[vault].nativePenalty, IncorrectNativePenalty());
+        require(activeAdapter[vault][adapter], InactiveAdapter());
+        // forge-lint: disable-next-item(unsafe-typecast) safe because msg.value == nativePenalty <= type(uint120).max.
+        if (msg.value > 0) vaultData[vault].accruedNativePenalty += uint120(msg.value);
+        require(vaultData[vault].canAllocateFromIdle, CannotDeallocate());
```
peyha, line 133:
```suggestion
        require(vaultData[vault].canAllocateFromIdle, CannotDeallocate());
        // forge-lint: disable-next-item(unsafe-typecast) safe because msg.value == nativePenalty <= type(uint120).max.
        if (msg.value > 0) vaultData[vault].accruedNativePenalty += uint120(msg.value);
```
nitpick

src/periphery/README.md
```diff
@@ -0,0 +1,24 @@
...
+`BluePublicAllocator` lets anyone (not only the vault's allocators) reallocate a Vault V2's liquidity between Morpho Blue markets reached through the `MorphoMarketV1AdapterV2`, within the vault's caps, and the public allocator's own limits.
+Vault allocators configure which adapters are active, which markets can be deallocated from, each market's absolute cap, and whether idle assets can be allocated publicly.
+Public callers must pay the vault's native penalty on each `reallocate` or `allocateFromIdle` call.
+
+## Whitelist Receive Shares Gate
+
+`WhitelistReceiveSharesGate` is an implementation of the vault's `receiveSharesGate`.
+It restricts who can receive vault shares, including shares minted on deposits/mints and shares received by transfer.
+A whitelisted account can still hold shares on behalf of non-whitelisted accounts, so this gate does not fully restrict who can access the vault's payoff.
+If an account transfers shares away, for example into another protocol, it may be unable to receive them back if it is un-whitelisted later.
```
QGarchery:
```suggestion
If an account transfers shares to another account, for example into another protocol, it may be unable to receive them back if that account is not whitelisted later.
```

peyha (approved):

QGarchery (approved):

peyha (approved):

adhusson (commented):
LGTM

PR #955, opened 2026-07-28
[midnight adapter] feat: sell dutch auction

peyha:
Let the vault sell by letting allocators sign auction offers which will implement a dutch auction (decaying price/decreasing rate)


peyha:
outdated

PR #965, opened 2026-08-11
[public allocator] docs: forceDeallocatePenalty

MathisGD:
https://cantina.xyz/code/b6e917be-dc53-4f66-b58e-406e4d1be2fe/findings/15


src/periphery/blue-public-allocator/BluePublicAllocator.sol
```diff
@@ -13,6 +13,7 @@ import {MarketParams} from "../../../lib/morpho-blue/src/interfaces/IMorpho.sol"
 /// @dev The vault's allocators can manage the public allocators' settings.
 /// @dev Each reallocate and allocateFromIdle call costs a penalty in native currency, set per vault by the allocators.
 /// The penalty is accrued per vault and can be claimed by the vault's allocators.
+/// @dev The vault's forceDeallocatePenalty is ignored by this contract.
```
peyha, line 16: feel like this comment is not so clear
we could be more explicit with something like

```suggestion
/// @dev If the vault's forceDeallocatePenalty is high while the public allocator's penalty is low, it could be cheaper to deallocate via the public allocator
```

MathisGD: I don't want to list all the specific scenarios but instead just point at the fact that this is ignored => it invalidates a bunch of issues and makes people think about what they need to care about wrt penalties. wdyt?

peyha: I don't think we should specify scenarios neither but the current version looks very blurry, I'm not sure people will grasp the implication on forcedeallocate

PR #968, opened 2026-08-13
[public allocator]: require non zero cap for allocation

peyha:


bhargavbh (commented):
is this mainly gas optimisation and adding the zero check?
IIUC there is no behaviour change. It now reverts on a zero allocation with zero allocateCap, which previously would be a no-op.

MathisGD (approved):

bhargavbh (approved):

adhusson (commented):
not sure about the point (only change is that it creates a noop check) but LGTM

MathisGD:
full context [here](https://morpholabs.slack.com/archives/C08A40KKN9W/p1786629201885609?thread_ts=1786548352.422749&cid=C08A40KKN9W)

PR #988, opened 2026-09-07
[midnight adapter] durations in adapter

adhusson:


adhusson:
closing in favor of durations in factory

PR #989, opened 2026-09-08
[midnight adapter] hardcoded duration

adhusson:


adhusson:
closing in favor of durations given in factory constructor

PR #991, opened 2026-09-09
[midnight adapter] eager loss realization

adhusson:
(ignore the branch name)

### DIFF
--- a/src/adapters/MidnightAdapter.sol
+++ b/src/adapters/MidnightAdapter.sol
@@ -51,6 +51,7 @@ contract MidnightAdapter is IMidnightAdapter {
     /* MANAGEMENT */
 
     address public skimRecipient;
+    bool public skipBufferCheck;
     mapping(address subRatifier => bool) public isSubRatifier;
 
     /* ACCOUNTING */
@@ -199,6 +200,12 @@ contract MidnightAdapter is IMidnightAdapter {
         timelocked();
         abdicated[selector] = true;
         emit Abdicate(selector);
+    }
+
+    function setSkipBufferCheck(bool newSkipBufferCheck) external {
+        timelocked();
+        skipBufferCheck = newSkipBufferCheck;
+        emit SetSkipBufferCheck(newSkipBufferCheck);
     }
 
     function setSkimRecipient(address newSkimRecipient) external {
@@ -389,16 +396,19 @@ contract MidnightAdapter is IMidnightAdapter {
         require(msg.sender == midnight, NotMidnight());
         require(seller == address(this), NotSelf());
 
+        uint256 vaultTotalAssetsBefore = IVaultV2(parentVault).totalAssets();
         int256 change = updateMarket(marketId, market, 0, 0);
 
         IVaultV2(parentVault).deallocate(address(this), abi.encode(ids(market), change), sellerAssets);
 
-        uint256 vaultRealAssetsAfter = IERC20(asset).balanceOf(parentVault);
-        uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
-        for (uint256 i = 0; i < adaptersLength; i++) {
-            vaultRealAssetsAfter += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
+        if (!skipBufferCheck) {
+            uint256 vaultRealAssetsAfter = IERC20(asset).balanceOf(parentVault);
+            uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
+            for (uint256 i = 0; i < adaptersLength; i++) {
+                vaultRealAssetsAfter += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
+            }
+            require(vaultRealAssetsAfter >= vaultTotalAssetsBefore, BufferTooLow());
         }
-        require(vaultRealAssetsAfter >= IVaultV2(parentVault).totalAssets(), BufferTooLow());
 
         // forge-lint: disable-next-item(unsafe-typecast) change <= 0 when no credit is bought.
         emit Sell(marketId, sellerAssets, uint256(-change));
@@ -461,7 +471,6 @@ contract MidnightAdapter is IMidnightAdapter {
         while (count < durationsLength && timeToMaturity >= packedDurations.get(count)) count++;
     }
 
-    /// @dev Liquidation cursors are omitted from collateral ids.
     function ids(Market memory market) public view returns (bytes32[] memory) {
         uint256 durationsCount = _maturities[market.maturity].durationCount;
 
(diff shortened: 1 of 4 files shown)


src/adapters/MidnightAdapter.sol
```diff
@@ -473,71 +386,70 @@ contract MidnightAdapter is IMidnightAdapter {
...
-            .deallocate(address(this), abi.encode(ids(market), -int256(netCreditDecrease)), sellerAssets);
-
-        uint256 loss = adapterTotalAssetsBefore.zeroFloorSub(uint256(totalAssets) + sellerAssets);
-        uint256 consumed = loss.zeroFloorSub(vaultBuffer);
-        if (consumed > 0) {
-            skipBufferAllowance -= consumed;
-            emit ConsumeSkipBufferAllowance(consumed);
+            vaultRealAssetsAfter += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
         }
+        require(vaultRealAssetsAfter >= IVaultV2(parentVault).totalAssets(), BufferTooLow());
```
adhusson: I removed the "allowance" check where we measure how much the vault has lost because i don't know how to make it work with nested sells, which can be triggered by the buyer's callback.

src/adapters/MidnightAdapter.sol
```diff
@@ -125,6 +117,7 @@ contract MidnightAdapter is IMidnightAdapter {
     }
 
     function isRatified(Offer memory offer, bytes memory data, address taker) external view returns (bytes32) {
+        require(IVaultV2(parentVault).firstTotalAssets() != 0, VaultNotAccrued());
```
adhusson, line 121: note this is not great at all since you can't take an adapter offer unless the vault was pinged first

src/adapters/MidnightAdapter.sol
```diff
@@ -285,61 +269,24 @@ contract MidnightAdapter is IMidnightAdapter {
...
+            bytes32 marketId = marketIds[i];
+            MarketData memory marketData = _markets[marketId];
+            uint256 newNetCredit;
+            if (marketData.lossFactor == IMidnight(midnight).lossFactor(marketId)) {
+                newNetCredit = marketData.netCredit;
+            } else {
+                // updatePositionView only reads market.maturity.
+                Market memory market;
+                market.maturity = marketData.maturity;
+                newNetCredit = currentNetCredit(marketId, market);
```
adhusson, line 292: saves 13k gas per market compared to

```solidity
newNetCredit = currentNetCredit(marketId, IMidnight(midnight).toMarket(marketId));
```

src/adapters/MidnightAdapter.sol
adhusson: not anymore

src/adapters/interfaces/IMidnightAdapter.sol
```diff
@@ -7,15 +7,18 @@ import {Market, Offer} from "lib/midnight/src/interfaces/IMidnight.sol";
 import {IBuyCallback, ISellCallback} from "lib/midnight/src/interfaces/ICallbacks.sol";
 import {IRatifier} from "lib/midnight/src/interfaces/IRatifier.sol";
 
-struct MaturityData {
+struct MarketData {
     uint128 netCredit;
-    uint120 growth;
-    uint8 durationCount;
+    uint128 lossFactor;
```
MathisGD: not used

src/adapters/MidnightAdapter.sol
```diff
@@ -389,134 +377,131 @@ contract MidnightAdapter is IMidnightAdapter {
...
     }
 
     /* INTERNAL FUNCTIONS */
 
-    function currentNetCredit(bytes32 marketId) internal view returns (uint256) {
-        return
-            IMidnight(midnight).credit(marketId, address(this))
-                - IMidnight(midnight).pendingFee(marketId, address(this));
+    function currentNetCredit(bytes32 marketId, Market memory market) internal view returns (uint128) {
+        (uint128 credit, uint128 pendingFee,) = IMidnight(midnight).updatePositionView(market, marketId, address(this));
```
MathisGD, line 456: you didn't do the skip if lossFactor didn't change optimization on purpose?

src/adapters/MidnightAdapter.sol
```diff
@@ -123,11 +123,11 @@ contract MidnightAdapter is IMidnightAdapter {
     }
 
     function isRatified(Offer memory offer, bytes memory data, address taker) external view returns (bytes32) {
+        require(!IMidnight(midnight).liquidationLocked(IdLib.toId(offer.market), address(this)), SellInProgress());
         // Collaterals will be checked through vault ids.
```
MathisGD:
```suggestion
        // Collaterals and duration will be checked through vault ids.
```

src/adapters/MidnightAdapter.sol
```diff
@@ -238,29 +238,22 @@ contract MidnightAdapter is IMidnightAdapter {
...
-
-    function withdrawToVaultInternal(Market memory market, uint256 withdrawnAssets, uint256 pendingBoughtNetCredit)
-        internal
-    {
         bytes32 marketId = IdLib.toId(market);
-        require(!IMidnight(midnight).liquidationLocked(marketId, address(this)), SellInProgress());
+        require(
+            IVaultV2(parentVault).isAllocator(msg.sender) || IVaultV2(parentVault).isSentinel(msg.sender),
+            NotAuthorized()
+        );
```
MathisGD: why not permissionless?

src/adapters/MidnightAdapter.sol
adhusson: i used to do it in `realAssets` (in `currentNetCredit` it would be too aggressive sometimes you check after a credit change unrelated to a loss)

```solidity
if (marketId == overridenMarketId) {
    newNetCredit = overridenMarketNetCredit;
if (marketData.lossFactor == IMidnight(midnight).lossFactor(marketId)) {
    newNetCredit = marketData.netCredit;
} else {
    <currentNetCredit>
}
```

but it didn't seem worth the extra branch & field (the one i forgot to remove).

I checked gas diff, for each market its 8k saved on cold, 4k saved on warm, and +1.5k if the loss factor did change (which is rarer). 

wdyt? still not sure it's worth the extra branch.

MathisGD: indeed it's an extra branch but with a very simple logic, sounds ok to me. we can re-add later too

src/adapters/MidnightAdapter.sol
adhusson: lost in a merge

src/adapters/MidnightAdapter.sol
MathisGD: let's merge this and re-add later

PR #992, opened 2026-09-09
[midnight adapter] incentivize loss realization

adhusson:


src/adapters/MidnightAdapter.sol
```diff
@@ -235,10 +235,6 @@ contract MidnightAdapter is IMidnightAdapter {
 
     function withdrawToVault(Market memory market, uint256 withdrawnAssets) external {
```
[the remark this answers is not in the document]
adhusson: it's an exposure without benefits

src/periphery/midnight-adapter-loss-realization-funder/MidnightAdapterLossRealizationFunder.sol
```diff
@@ -0,0 +1,77 @@
...
+            require(markets[i].loanToken == asset, IMidnightAdapter.LoanAssetMismatch());
+            IMidnightAdapter(adapter).withdrawToVault(markets[i], 0);
+            marketIds[i] = IdLib.toId(markets[i]);
+        }
+
+        uint256 loss = adapterAssets - IMidnightAdapter(adapter).realAssets();
+        if (loss > 0 && minimumLossBeforeIncentive > 0 && loss >= minimumLossBeforeIncentive) {
+            uint256 paid = incentive;
+            // forge-lint: disable-next-item(arbitrary-send-eth) caller chooses the incentive receiver.
+            (bool success,) = receiver.call{value: paid}("");
```
peyha, line 68: as a liquidator, if I see a large liquidatable position with bad debt, I can repeat the following actions:
1. Liquidate part of the position (just enough so that loss >= minimumLossBeforeIncentive)
2. realizeLoss to earn the incentive
not a huge fan of this, a possible fix would be to scale the incentive by the loss

peyha: this gives a pretty cool guarantee, with a given amount of incentive, we know how much loss it can cover

adhusson: It is possible to partially liquidate, but it is not possible to partially realize bad debt. The liquidator will realize the entire bad debt associated with a borrower even when liquidating 0.

That said it makes sense to increase the payment with the loss as realizing becomes more important when the loss increases.

adhusson: https://github.com/morpho-org/vault-v2/pull/992/commits/f0031f2aa41d4aead904535e23ce6d983f336d27

adhusson:
closing as we are going to eager loss realization

PR #994, opened 2026-09-10
[midnight adapter] less accrual cache

adhusson:
up to +130k gas (cold) if there are 50 future maturities. +10k if warm. Cached for the current timestamp.


MathisGD (approved):
net improvement, let's merge that

PR #995, opened 2026-09-11
[midnight adapter] WAD-scaled growth

adhusson:


adhusson:
closing as it only applies to lazy loss realization (https://github.com/morpho-org/vault-v2/pull/786/changes/fb82768f5605bd5e699759e86a673467a925804e)

PR #1000, opened 2026-09-17
[midnight adapter] Dutch auction + measure loss by book value

adhusson:


src/adapters/MidnightAdapter.sol
```diff
@@ -253,6 +297,27 @@ contract MidnightAdapter is IMidnightAdapter {
         emit WithdrawToVault(marketId, withdrawnAssets, uint256(-change));
     }
 
+    /// @dev Auction credit after all should have been repaid.
+    /// @dev Ignores the lossAllowance mapping.
+    function withdrawFor(Market memory market, uint256 units, address receiver, bytes memory data) external {
+        bytes32 marketId = IdLib.toId(market);
+        uint256 start = market.maturity + TIME_TO_MAX_LIF;
+        uint256 duration = start < auctionDurationChangedAt ? previousAuctionDuration : auctionDuration;
+        uint256 sellerAssets = units.mulDivUp(duration.zeroFloorSub(block.timestamp - start), duration);
```
adhusson: The auction is inside the adapter, not in a subratifier, so that it does not depend on loss allowance. Having per-ratifier loss allowance is not possible because `onSell` does not get the ratifier as an argument.

src/adapters/MidnightAdapter.sol
```diff
@@ -253,6 +297,27 @@ contract MidnightAdapter is IMidnightAdapter {
...
+
+        (uint128 credit,,) = IMidnight(midnight).updatePositionView(market, marketId, address(this));
+        withdrawToVault(market, MathLib.min(IMidnight(midnight).withdrawable(marketId), credit));
```
adhusson, line 285: Here we can withdraw first, but in onSell we can't.

src/adapters/MidnightAdapter.sol
```diff
@@ -253,6 +297,27 @@ contract MidnightAdapter is IMidnightAdapter {
         emit WithdrawToVault(marketId, withdrawnAssets, uint256(-change));
     }
 
+    /// @dev Auction credit after all should have been repaid.
+    /// @dev Ignores the lossAllowance mapping.
+    function withdrawFor(Market memory market, uint256 units, address receiver, bytes memory data) external {
+        bytes32 marketId = IdLib.toId(market);
+        uint256 start = market.maturity + TIME_TO_MAX_LIF;
```
adhusson: wait until all normal markets have realistically been repaid

src/adapters/MidnightAdapter.sol
```diff
@@ -253,6 +297,27 @@ contract MidnightAdapter is IMidnightAdapter {
...
+        uint256 duration = start < auctionDurationChangedAt ? previousAuctionDuration : auctionDuration;
+        uint256 sellerAssets = units.mulDivUp(duration.zeroFloorSub(block.timestamp - start), duration);
+
+        (uint128 credit,,) = IMidnight(midnight).updatePositionView(market, marketId, address(this));
+        withdrawToVault(market, MathLib.min(IMidnight(midnight).withdrawable(marketId), credit));
+        IWithdrawForCallback(msg.sender).onWithdrawFor(data);
```
adhusson, line 287: it may be simpler for integration to not have a callback and instead do this:

- a transient `withdrawnMarketId` is set when a `withdrawToVault` call empties the withdrawable
- caller of withdrawFor does `withdrawToVault(max);liquidate();withdrawFor()`

but the code is less simple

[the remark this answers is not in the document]
adhusson: same result by accruing before the call to `onWithdrawFor`

peyha: can't you also require withdrawable to be 0 at the start of `withdrawFor` ?

src/adapters/MidnightAdapter.sol
```diff
@@ -256,6 +274,28 @@ contract MidnightAdapter is IMidnightAdapter {
         emit WithdrawToVault(marketId, withdrawnAssets, uint256(-change));
     }
 
+    /// @dev Auction credit after all should have been repaid.
+    /// @dev Ignores the lossAllowance mapping.
+    function withdrawFor(Market memory market, uint256 units, address receiver, bytes memory data) external {
```
peyha: I don't find the name of the function very clear
prop: `auctionCredit`

src/adapters/MidnightAdapter.sol
```diff
@@ -429,20 +469,13 @@ contract MidnightAdapter is IMidnightAdapter {
...
-            uint256 vaultRealAssetsAfter = IERC20(asset).balanceOf(parentVault) + sellerAssets;
-            uint256 adaptersLength = IVaultV2(parentVault).adaptersLength();
-            for (uint256 i = 0; i < adaptersLength; i++) {
-                vaultRealAssetsAfter += IAdapter(IVaultV2(parentVault).adapters(i)).realAssets();
-            }
-
-            (overridenMarketId, overridenMarketNetCredit) = (bytes32(0), 0);
-            require(vaultRealAssetsAfter >= vaultTotalAssetsBefore, BufferTooLow());
-        }
+        uint256 discountFactor =
```
peyha: growth is rounded down so `discountFactor` is rounded up meaning `loss` is also rounded up
I think that's the right rounding direction but we should make sure that losses cannot be inflated/multiplied

src/adapters/MidnightAdapter.sol
adhusson: With the transient version the withdrawable will not be zero at the start; instead we require `withdrawnMarketId == marketId` and this shows a full withdraw was done in this tx before the liquidation happened.

src/adapters/interfaces/IMidnightAdapter.sol
```diff
@@ -7,6 +7,9 @@ import {Market, Offer} from "lib/midnight/src/interfaces/IMidnight.sol";
 import {IBuyCallback, ISellCallback} from "lib/midnight/src/interfaces/ICallbacks.sol";
 import {IRatifier} from "lib/midnight/src/interfaces/IRatifier.sol";
 
+uint256 constant AUCTION_DELAY = 1 days;
+uint256 constant AUCTION_DURATION = 3 days;
```
albert-llimos, line 11: Are we set on hardcoding this? Allowing to change them on the fly seems unnecessary but maybe setting them in the constructor?

adhusson: constructor works, but it's simpler if there is no choice to make and we think those values work for everyone

src/adapters/MidnightAdapter.sol
adhusson: it's not on this branch anymore but on `decrease-per-time` and `loss-per-second-shortcut` the shortfall is still computed like this, and matches exactly the way realAssets computes the difference (discountFactor * netcredit).

src/adapters/interfaces/IMidnightAdapter.sol
albert-llimos: Agreed on hardcoded values on the call. Resolving.

